PR #6, opened 2016-09-02
Make $capabilities initialize param optional

kaloyan-raev:
Eclipse Che does not pass ClientCapabilities in InitializeParams. This
works well with other language server, but not with this one.

The issue is resolved by making the $capabilities param optional.


felixfbecker:
Eclipse Che behaves incorrectly here, `capabilitities` is not optional:
https://github.com/Microsoft/language-server-protocol/blob/master/protocol.md#initialize-request

It can be empty, but it must be defined

kaloyan-raev:
OK. I prepared this PR for Che: https://github.com/eclipse/che/pull/2314

PR #10, opened 2016-09-05
Support document formatting #8

### LINKED ISSUE #8: formatDocument
It can use the PHPParser's pretty printer

mniewrzal:
Closes #8


src/Server/TextDocument.php
```diff
@@ -124,4 +126,25 @@ private function updateAst(string $uri, string $content)
...
+     * The document formatting request is sent from the server to the client to format a whole document.
+     *
+     * @param TextDocumentIdentifier $textDocument The document to format
+     * @param FormattingOptions $options The format options
+     * @return TextEdit[]
+     */
+    public function formatting(TextDocumentIdentifier $textDocument, FormattingOptions $options)
+    {
+        $nodes = $this->asts[$textDocument->uri];
+        if (empty($nodes)){
```
felixfbecker: Space missing before `{`

src/Server/TextDocument.php
```diff
@@ -124,4 +126,25 @@ private function updateAst(string $uri, string $content)
...
+            return [];
+        }
+        $prettyPrinter = new PrettyPrinterStandard();
+        $edit = new TextEdit();
+        $edit->range = new Range(new Position(0, 0), new Position(PHP_INT_MAX, PHP_INT_MAX));
```
felixfbecker: Is this valid?

src/Server/TextDocument.php
```diff
@@ -2,7 +2,7 @@
 
 namespace LanguageServer\Server;
 
-use PhpParser\{Error, Comment, Node, ParserFactory, NodeTraverser, Lexer};
+use PhpParser\{Error, Comment, Node, ParserFactory, NodeTraverser, Lexer, PrettyPrinter\Standard as PrettyPrinterStandard};
```
felixfbecker: Can you put that in a seperate `use` statement please?

felixfbecker:
This needs a unit test
See https://github.com/felixfbecker/php-language-server/blob/master/tests/Server/TextDocumentTest.php

src/Server/TextDocument.php
mniewrzal: I didn't found anything about it in protocol spec. Most probably it will depend on client implementation. It works for VS Code but if you feel uncomfortable with it I will reorganize code to store opened documents to calculate end position or traverse to last ast node and get end position from it.

felixfbecker: I think it's okay as-is. Other options would be to use `$nodes[count($nodes) - 1]->getAttribute('fileEndPos')` or introduce a new AST object that holds the statements but also the raw file content.

src/Server/TextDocument.php
```diff
@@ -124,4 +127,25 @@ private function updateAst(string $uri, string $content)
...
+     */
+    public function formatting(TextDocumentIdentifier $textDocument, FormattingOptions $options)
+    {
+        $nodes = $this->asts[$textDocument->uri];
+        if (empty($nodes)) {
+            return [];
+        }
+        $prettyPrinter = new PrettyPrinter();
+        $edit = new TextEdit();
+        $edit->range = new Range(new Position(0, 0), new Position(PHP_INT_MAX, PHP_INT_MAX));
```
felixfbecker: @dbaeumer is it okay to do this to update the whole document?

fixtures/Format.php
```diff
@@ -0,0 +1,20 @@
+<?php
```
felixfbecker: Can you please lowercase these file names as they are not class files?

felixfbecker:
Looking forward to using this while developing on the language server itself! Always eat your own dog food 😄

felixfbecker:
LGTM

src/Server/TextDocument.php
dbaeumer: @felixfbecker yes and no. It is always best if an extension provides a minimal edit set since it will preserve the most markers attached to the text. We have code in place that runs a diff in the whole document gets replaced to compute some more minimal diff. However this might be suboptimal as well. So the more fine grain your edits are the better.

felixfbecker: @dbaeumer Of course a minimal text edit is better, but the default implementation of PHPParser's PrettyPrinter will just spit out the new whole new content. We may customize the behavior later, but for now we just wanted to get the feature done. My question was only about wether it is okay to use a very big number (PHP_INT_MAX) to make it update the whole document (even though the EOF offset is much smaller) or if that is not protocol-conform and causes problems.

Issue #12, opened 2016-09-05
Add linting


felixfbecker:
I don't know what is the best linter for PHP and how to configure it?
Needs to be integrated in CI

MadHed:
I've been using https://github.com/squizlabs/PHP_CodeSniffer at work.
It's extendable with custom validation rules and already contains rules for PSR-2.
It's available as a composer package and actively maintained.

Overall very nice, haven't integrated it with a CI solution, howver.

kaloyan-raev:
How about using the [PHP Mess Detector](https://phpmd.org) for additional static code analysis?

felixfbecker:
What I desire is a linter, that can be configured through a config file and defaults to some coding best practices, that can be installed with Composer, run in CI and has an editor plugin for VS Code

kaloyan-raev:
As far as I am familiar there is no one tool to rule them all. http://phpqatools.org/ is a nice collection of popular tools that are used in CI. Some of them like Code Sniffer and Mess Detector are also often integrated in the editor/IDE.

I can imagine that the language server look for the specific configuration files in the project and if found, activate the corresponding linting tool. For example, if there is a `phpcs.xml` file then the language server will activate use Code Sniffer to report additional diagnostics. Of course, the client should be able to override this default behavior.

felixfbecker:
@kaloyan-raev This issue is not for adding linting to the LS, it is for linting this repository (for PRs etc) 😄

kaloyan-raev:
OK :)

ADmad:
> ..and has an editor plugin for VS Code

This phpcs plugin works just fine https://marketplace.visualstudio.com/items?itemName=ikappas.phpcs

PR #15, opened 2016-09-08
Variable symbols

levan-g:
We should exclude variable symbols that are defined in function and method body.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 .DS_Store
 .vscode
+.idea
 vendor/
 composer.lock
--- a/src/SymbolFinder.php
+++ b/src/SymbolFinder.php
@@ -44,8 +44,8 @@ class SymbolFinder extends NodeVisitorAbstract
     {
         $class = get_class($node);
         if (!isset(self::NODE_SYMBOL_KIND_MAP[$class])) {
-        return;
-    }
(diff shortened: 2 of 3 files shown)


felixfbecker:
Should we? I remember that TypeScript language server once behaved that way, and they changed the behavior:
https://github.com/Microsoft/TypeScript/issues/8218

levan-g:
I have latest version of typescript and it does not show symbols inside functions. 

From my point of view it's better to hide symbols inside functions.
You can open some big php file and you'll see that it's quite uncomfortable to navigate to other method in class when you have variable symbols.

levan-g:
Alternatively we can add option to enable/disable this behavior.

felixfbecker:
I double checked and you are right, we should follow TS' behavior here. A setting may be nice for a future update, but low priority.

felixfbecker:
Could you rebase this on master (so it only includes the actual code changes)?

levan-g:
yes.

felixfbecker:
To prevent this in the future, do you have a recommendation for #12 ?

levan-g:
I was unable to rebase the branch so I created another pull request https://github.com/felixfbecker/php-language-server/pull/16 .
you can close this one.

PR #31, opened 2016-09-18
Global symbol search

MadHed:
Closes #31 

Parsing the whole php-language-server project takes about 10 seconds on my machine.
The php-cli process takes up about 180MB of ram with all symbol information parsed.

Changes made:
* Added support for client:window/* commands
* Added support for server:workspace/* commands
* Refactored symbol database into own set of classes. (Was in Server/TextDocument before)
* Disabled memory limit
* ProtocolStreamReader was blocking the event loop, made non-blocking
* Find all .php files on startup and start parsing one at a time with Loop/nextTick

To do:
* [x] Fix old tests
* [x] Add new tests
* [x] Reduce memory usage. Most of it is the AST, maybe we can throw it away after parsing,
* [x] Figure out solution for blocking STDIN. The current one is a hack. This is a windows limitation, see [https://bugs.php.net/bug.php?id=34972](https://bugs.php.net/bug.php?id=34972)
* [x] Extract URI generation into utils.php
* [x] Investigate escaping Parser Exception
* [x] Open issue at PhpParser regarding Lexer exceptions
* [x] Open issue at php-intellisense regarding OS switch
* [x] Upgrade Sabre/event dependency (4.0.0)
* ~~Can use NameResolver for generating FQN containerNames~~


src/LanguageServer.php
```diff
@@ -69,11 +82,18 @@ public function __construct(ProtocolReader $reader, ProtocolWriter $writer)
      */
     public function initialize(string $rootPath, int $processId, ClientCapabilities $capabilities): InitializeResult
```
felixfbecker: I wasn't aware `$rootPath` can be null, need to remove the type hint then and adapt the `@param`. Or even better, move it to the end as a parameter and give it default value of `null`. There should also be a test for this.

src/LanguageServer.php
```diff
@@ -100,4 +120,42 @@ public function exit()
...
+        $files = new \RegexIterator($ite, '/^.+\.php$/i', \RegexIterator::GET_MATCH);
+        $fileList = array();
+        foreach($files as $file) {
+            $fileList = array_merge($fileList, $file);
+        }
+        
+        $processFile = function() use (&$fileList, &$processFile, &$rootPath){
+            if ($file = array_pop($fileList)) {
+                
+                $uri = 'file://'.(substr($file, -1) == '/' || substr($file, -1) == '\\' ? '' : '/').str_replace('\\', '/', $file);
```
felixfbecker: We should make a helper function for this, will need it more often. I am for adding a `utils.php` with plain functions that are autoloaded through composer.json `files` property

src/PhpDocument.php
```diff
@@ -0,0 +1,112 @@
...
+
+namespace LanguageServer;
+
+use \LanguageServer\Protocol\{Diagnostic, DiagnosticSeverity, Range, Position, SymbolKind, TextEdit};
+
+use PhpParser\{Error, Comment, Node, ParserFactory, NodeTraverser, Lexer, Parser};
+use PhpParser\PrettyPrinter\Standard as PrettyPrinter;
+use PhpParser\NodeVisitor\NameResolver;
+
+class PhpDocument
```
felixfbecker: 👍 for this class, wanted to do that at one point aswell

src/PhpDocument.php
```diff
@@ -0,0 +1,112 @@
...
+     * @return void
+     */
+    public function updateAst(string $content)
+    {
+        $stmts = null;
+        try {
+            $stmts = $this->parser->parse($content);
+        }
+        catch(Error $e) {
+            // Parser still throws errors. e.g for unterminated comments
```
felixfbecker: Is this really true? If yes, isn't that a bug in PHPParser?
Probably also shouldn't catch all `Error`s but only `Exception` or PHPParser's Exception type

src/Project.php
```diff
@@ -0,0 +1,52 @@
...
+
+namespace LanguageServer;
+
+use PhpParser\{Error, Comment, Node, ParserFactory, NodeTraverser, Lexer};
+use PhpParser\PrettyPrinter\Standard as PrettyPrinter;
+use PhpParser\NodeVisitor\NameResolver;
+
+class Project
+{
+    private $documents;
```
felixfbecker: Need some `@var` annotations here and document this is an associative array of URI => document

src/ProtocolStreamReader.php
```diff
@@ -30,33 +30,32 @@ public function __construct($input)
...
-                    case ParsingMode::HEADERS:
-                        if ($this->buffer === "\r\n") {
-                            $this->parsingMode = ParsingMode::BODY;
-                            $this->contentLength = (int)$this->headers['Content-Length'];
-                            $this->buffer = '';
-                        } else if (substr($this->buffer, -2) === "\r\n") {
-                            $parts = explode(':', $this->buffer);
-                            $this->headers[$parts[0]] = trim($parts[1]);
-                            $this->buffer = '';
+            $c = fgetc($this->input);
```
felixfbecker: Could you commend a bit on this change?
As I understand [`Loop\addReadStream`](http://sabre.io/event/loop/), it calls the callback as soon as the stream has data in the buffer using `stream_select` under the hood. So as long as there is no data, it will not block. When the callback is called, it could be that there is more than one character in the buffer, so I call `fgetc` (blocking) until the buffer is empty again. I have to admit I was/am not familiar at all with what PHP stream read functions are suitable here as I never worked with them. I used `fgetc` because I wanted to be 100% sure I don't loose any data and the other functions seemed to only read a chunk of a specific size (while here `Loop\addReadStream` should already take care of the chunking)

src/Server/TextDocument.php
```diff
@@ -79,7 +57,7 @@ public function documentSymbol(TextDocumentIdentifier $textDocument): array
      */
     public function didOpen(TextDocumentItem $textDocument)
     {
-        $this->updateAst($textDocument->uri, $textDocument->text);
+        $this->project->getDocument($textDocument->uri)->updateAst($textDocument->text);
```
felixfbecker: In case this proves to be a long blocking call, we could also think about queuing that to the end of the event loop and returning a promise.

src/Server/TextDocument.php
```diff
@@ -24,29 +21,18 @@
...
-    private $parser;
-
-    /**
-     * A map from file URIs to ASTs
-     *
-     * @var \PhpParser\Stmt[][]
-     */
-    private $asts;
-
-    /**
```
felixfbecker: 👍 for separation of concerns

src/Server/Workspace.php
```diff
@@ -0,0 +1,73 @@
...
+
+    /**
+     * A notification sent from the client to the server to signal the change of configuration settings.
+     *
+     * @param The actual changed settings
+     * @return void
+     */
+    public function didChangeConfiguration($settings)
+    {
+    }
```
felixfbecker: Is there a reason you implemented empty methods here? If they are not implemented the JSON RPC dispatcher should just return a METHOD_NOT_IMPLEMENTED error

src/SymbolFinder.php
```diff
@@ -35,31 +35,56 @@ class SymbolFinder extends NodeVisitorAbstract
...
+            array_push($this->nameStack, $containerName);
+        }
+
         $class = get_class($node);
         if (!isset(self::NODE_SYMBOL_KIND_MAP[$class])) {
             return;
         }
 
+        // if we enter a method or function, increase the function counter
+        if ($class === Node\Stmt\Function_::class || $class === Node\Stmt\ClassMethod::class) {
```
felixfbecker: Why not `instanceof` here?

src/SymbolFinder.php
```diff
@@ -72,13 +97,19 @@ public function enterNode(Node $node)
...
 
     public function leaveNode(Node $node)
     {
-        $this->containerName = null;
+        array_pop($this->nodeStack);
+        array_pop($this->nameStack);
+
+        // if we leave a method or function, decrease the function counter
+        $class = get_class($node);
+        if ($class === Node\Stmt\Function_::class || $class === Node\Stmt\ClassMethod::class) {
```
felixfbecker: Also here, why not `instanceof`?

src/LanguageServer.php
```diff
@@ -100,4 +120,42 @@ public function exit()
...
+                
+                $numFiles = count($fileList);
+                if (($numFiles % 100) == 0) {
+                    $this->client->window->logMessage(3, $numFiles.' PHP files remaining.');
+                }
+                
+                $this->project->getDocument($uri)->updateAst(file_get_contents($file));
+                
+                Loop\nextTick($processFile);
```
felixfbecker: I think this should be `setTimeout`. `nextTick` will queue the callback before any I/O callbacks. `setTimeout` on the other hand will push it to the end of the queue, first executing all other callbacks (like the STDIN handling for example), then getting back to processing. See this note from the Node docs:

> This is not a simple alias to setTimeout(fn, 0), it's much more efficient. It runs before any additional I/O events (including timers) fire in subsequent ticks of the event loop.
> Note: the next tick queue is completely drained on each pass of the event loop before additional I/O is processed. As a result, recursively setting nextTick callbacks will block any I/O from happening, just like a while(true); loop.

The current way would mean that I/O is blocked (including STDIN), so the language server would become unresponsive until the indexing is finished.

src/PhpDocument.php
```diff
@@ -0,0 +1,112 @@
...
+    /**
+     * Returns symbols in this document filtered by query string.
+     *
+     * @param string $query The search query
+     * @return SymbolInformation[]
+     */
+    public function findSymbols(string $query)
+    {
+        return array_filter($this->symbols, function($symbol) use(&$query) {
+            return stripos($symbol->name, $query) !== false;
```
felixfbecker: Fuzzy searching would be nice for a future update

src/LanguageServer.php
```diff
@@ -100,4 +120,42 @@ public function exit()
...
         exit(0);
     }
+
+    /**
+     * Parses workspace files, one at a time.
+     *
+     * @param string $rootPath The rootPath of the workspace. Is null if no folder is open.
+     * @return void
+     */
+    private function indexProject(string $rootPath)
```
felixfbecker: I think you cannot pass `null` if you type hint it as `string`. It needs to be `string $rootPath = null` then.

src/LanguageServer.php
```diff
@@ -100,4 +120,42 @@ public function exit()
...
+        }
+        
+        $processFile = function() use (&$fileList, &$processFile, &$rootPath){
+            if ($file = array_pop($fileList)) {
+                
+                $uri = 'file://'.(substr($file, -1) == '/' || substr($file, -1) == '\\' ? '' : '/').str_replace('\\', '/', $file);
+                
+                $numFiles = count($fileList);
+                if (($numFiles % 100) == 0) {
+                    $this->client->window->logMessage(3, $numFiles.' PHP files remaining.');
```
felixfbecker: Would be nice to log the file name aswell

src/LanguageServer.php
MadHed: Ok. I didn't know the dispatcher figures out the correct parameter position, cool!

src/LanguageServer.php
MadHed: Shouldn't we check this at the calling site? Indexing a null path doesn't make sense, after all.

src/LanguageServer.php
MadHed: Sounds good to me. Will also add a test for this.

src/LanguageServer.php
MadHed: Ok. SetTimeout would also allow dynamic throttling later.

src/LanguageServer.php
felixfbecker: Yeah, you're right. I just went by the `@param` desc

src/PhpDocument.php
MadHed: I tried to 😀 maybe you have a link?

src/PhpDocument.php
MadHed: I found this weird as well. I had the extension crash multiple times while writing a block comment.
But you are right, we should probably catch this specific extension and add it as a parse error.

src/Project.php
MadHed: ok

src/Server/TextDocument.php
MadHed: Could very well happen for large, complicated source files.

src/Server/Workspace.php
MadHed: I will remove these, they were just for testing.

src/SymbolFinder.php
MadHed: No particular reason, just reused the $class variable. Instanceof would probably make the purpose clearer.

src/SymbolFinder.php
MadHed: No particular reason, just reused the $class variable. Instanceof would probably make the purpose clearer.

src/LanguageServer.php
MadHed: What do you mean? Log the name of the next file to be parsed?

src/PhpDocument.php
felixfbecker: I would not implement it myself but look for a library from composer that we can integrate (or improve so it can be integrated here).

src/PhpDocument.php
felixfbecker: We clearly set `['throwOnError' => true]` so Imo it should never throw. If it does, could you open an issue at PHPParser, posting the exception? We are using the 3.0 alpha so bugs can happen.

src/Server/TextDocument.php
felixfbecker: What do you think is "large" in that context? But I think I am talking BS here because the parser itself (for a single document) will always be synchronous, if we don't spawn a seperate process (which would probably be much more overhead especially on Windows).

src/LanguageServer.php
felixfbecker: Yeah something like
```
Parsing src\LanguageServer.php (1/100)
Parsing src\LanguageClient.php (2/100)
```

src/ProtocolStreamReader.php
MadHed: This is the tricky part... There are multiple problems with the way the stream is currently read. And one of those is going to be very hard to fix...

You are looping while `fgetc() !== false`, which tests only for EOF (stream closed). So the loop would go on forever, because it simply waits if no new data is available.

Normally we _could_ set the STDIN stream to nonblocking with `stream_set_blocking(STDIN, false)` Then `fgetc` would simply return an empty string if no new data is available instead of blocking.

But! There is a bug/weird behaviour in windows that makes it impossible to set STDIN to non-blocking.

We could maybe call `stream_select` with a zero timeout before **any** call to `fgetc` but I don't know the performance implications of this.

Otherwise, we would have to spawn a second process/thread to only read from STDIN and pass the data via a socket, which can be non-blocking on windows too.

MadHed:
> I like this a lot :)

Thanks :)

> Could you comment how large the project was that took 10 seconds (amount of files, SLOC)?

It was **this** project :) php-language-server with all dependencies included. I don't know the LOC but it has around 1300 Files.

> Regarding memory usage

I also think 180MB is fine, however there is room for improvement.

> Regarding blocking calls

Maybe I am just overthinking this. I will test on a linux machine tomorrow and report back.

mniewrzal:
Hi, 
This PR looks very nice :) I'm still trying to test is on Linux but for large projects (1k files) indexing is not finishing and for smaller project I cannot get results;) I will try to investigate this and I will let you know. 

I want just add a comment about memory consumption. Right now you are talking about relatively small project (1k-2k files). If amount of memory will grow linear then for larger project (e.g. [magento](https://github.com/magento/magento2) 20k+) this value will be extremely high. I think keeping AST for all files from workspace is unnecessary because most of the time such detailed information aren't needed. Small, light model (abstract layer like namespace/class/method) created from AST would be also more useful from code readability point of view.

src/ProtocolStreamReader.php
felixfbecker: You are right, it will wait for EOF, I thought it would check if the buffer still has data. Would it help if we instead used `fgets`? `stream_select` is used by `Loop\addReadStream`, but we need some way to only read until the buffer is empty... If it always reads until EOF we have a problem

Is the problem really a Windows limitation, not a PHP bug? Because how do all the other language server like the TS server do it then? They all talk through STDIO.

felixfbecker:
@mniewrzal Tests are failing currently, I don't expect it to work 😄

MadHed:
Hold your horses. 😄  I just pushed some fixes, so I can test tomorrow.

@mniewrzal I think most of the AST is of little interest to us, anyway. The only situation where the full AST is needed is code formatting, but the user expects this to be relatively slow, so we should be able to reparse in this case.

@felixfbecker I had to adjust the expected result for the testDocumentSymbol test. Both methods and properties should have their class as containerName, right?

I reverted the fgetc logic for now. It doesn't work on windows anymore (only parses a file after some message arrived) but if I understand the behaviour it should work now under linux. I will test this tomorrow...

src/LanguageServer.php
```diff
@@ -75,12 +75,12 @@ public function __construct(ProtocolReader $reader, ProtocolWriter $writer)
     /**
      * The initialize request is sent as the first request from the client to the server.
      *
-     * @param string $rootPath The rootPath of the workspace. Is null if no folder is open.
      * @param int $processId The process Id of the parent process that started the server.
      * @param ClientCapabilities $capabilities The capabilities provided by the client (editor)
+     * @param string $rootPath The rootPath of the workspace. Is null if no folder is open.
```
felixfbecker: The type here must be `string|null` (this is significant for the dispatcher)

src/ProtocolStreamReader.php
MadHed: In my latest commit I added a `return` after a full message has been parsed. This should fix the problem with `fgetc` as the control is returned to the mainloop where `stream_select` is called again.

The real bummer is that `stream_select` on windows behaves erratically with STDIN.
Even if I set timeout to zero, it will always block without data on the stream. :/

If I understand correctly, there exists some very kludgy workarounds for this problem. NodeJS probably implemented them, while PHP did not. :/

mniewrzal:
> That's why I asked how big the project is this was tested with. I think we could add to the PHPDocument class some properties that keep the nodes that are relevant to us. We could then set the $stmts to null and only reload it when we need it. This would also result in faster lookup than walking the whole AST every time.

AST node contains lot of redundant information. At the beginning collecting only some of nodes would be ok but when more functions will be added to LS and memory usage will grow then I believe nodes should be transform into something more optimal. Parsing single file is fast and in many places can be done on demand.

> Also see my comment about how we could limit the PHPDocument objects in memory and then either remove them or serialize and write them to a cache on disk, what do you think about this?

I agree. Mechanism for storing collected data should be added later. Not only to reduce amount of used memory but also to avoid parsing after every initialization. Just not sure at the moment what will be the best approach ;-)

felixfbecker:
What kind of structure would you propose? Just as a reminder, we also need to save stuff from the docblocks in addition to the node. We need
- FQN as key
- SymbolKind
- comment / description
- for functions, signature for signature help (parameters with type and description)

The protocol has structures to represent all these so it makes sense to convert the AST / PHPDocumentator objects to these structures directly and only save those.

mniewrzal:
Yes, using protocol structures would be easiest solution. Of course if Symbolinformation/SignatureInfromation/... can store all necessary data. Also its possible that the same information will be duplicated across this structures. At this level I think it will be suitable. For later I'm thinking more about one structure, simpler and much smaller than AST, easy to produce protocol structures from it. For advanced code assist  (type detection/type inference)  language server need to collect many things like:
- inheritance
- access modifiers
- phpDoc tags

But as I wrote this is future. Before that this PR will be great extension to language server capabilities :-)

src/LanguageServer.php
MadHed: Changed

src/LanguageServer.php
MadHed: Done, also added a memory diagnostic message

src/SymbolFinder.php
MadHed: Done

src/Server/Workspace.php
MadHed: Done

src/SymbolFinder.php
MadHed: Done

src/Project.php
MadHed: Done

src/LanguageServer.php
MadHed: Done

src/LanguageServer.php
MadHed: Done

src/LanguageServer.php
MadHed: Done

src/PhpDocument.php
MadHed: This is probably a misunderstanding. The comment indicates that this setting only prevents the parser from bailing out with an exception on the first parse error and instructs it to continue parsing anyway. The exception here seems to originate from the Lexer.

src/ProtocolStreamReader.php
MadHed: I opened a [PR for the sabre project](https://github.com/fruux/sabre-event/pull/40).
This doesn't fix the windows problem, however. Maybe we need an OS switch in the extension.js.

src/utils.php
```diff
@@ -0,0 +1,31 @@
...
+}
+
+/**
+ * Transforms an absolute file path into a URI as used by the language server protocol. 
+ *
+ * @param string $filepath
+ * @return string
+ */
+function pathToUri(string $filepath): string {
+    return 'file://'.($filepath[0] == '/' || $filepath[0] == '\\' ? '' : '/').str_replace('\\', '/', $filepath);
```
felixfbecker: I think this implementation is not correct. A file URI must _always_ start with `file:///` (triple slash). In your case, you are checking if the first char is a backslash, but on windows it could be a drive letter. Also, special chars need to be URL-encoded. There is a very good implementation I've been using on NPM: https://github.com/sindresorhus/file-url/blob/master/index.js#L17-L24

felixfbecker: Test: https://github.com/sindresorhus/file-url/blob/master/test.js

felixfbecker:
I wonder how we can test the Windows-blocking issue?

MadHed:
I did a quick test with communicating over a socket. This worked flawlessly. We should put an OS switch into the client side script and spawn the Server with an argument, telling it to connect to a socket.

felixfbecker:
But if this really is a Windows issue, not a PHP one, then how do all the other language servers solve this? TCP has a lot of overhead compared to STDIO.

MadHed:
I think it is kind of both, but I am really not sure. stdin on windows behaves differently than on *nix but php doesn't include a workaround to be able to treat it as a normal stream.
I think the other LS are using environments where these workarounds are implemented... but who knows.

[Loopback sockets don't seem to be so bad performance-wise after all.](http://stackoverflow.com/questions/10872557/how-slow-are-tcp-sockets-compared-to-named-pipes-on-windows-for-localhost-ipc)

I'm really out of Ideas here.

Personally, I didn't notice any difference in performance.

felixfbecker:
This SO answer states that you can read STDIN non-blocking in Windows on a line-by-line basis, which is fine for our usecase as we don't need to process individual chars, only full responses. Maybe we have to look at the implementation of `addReadStream`?

bin/php-language-server.php
```diff
@@ -10,6 +12,21 @@
     }
 }
 
-$server = new LanguageServer(new ProtocolStreamReader(STDIN), new ProtocolStreamWriter(STDOUT));
+if (count($argv) >= 3 && $argv[1] == '--tcp') {
+    $address = $argv[2];
+    $socket = stream_socket_client('tcp://' . $address, $errno, $errstr);
```
felixfbecker: This means the IDE has to open a socket on its side, right?

MadHed: Yes. extension.js opens a listen socket at 127.0.0.1:random_port and immediately stops listening when the language server connects.

MadHed:
> This SO answer states that you can read STDIN non-blocking in Windows on a line-by-line basis.
The problem ist that `stream_select` blocks on windows using `STDIN` when no data is available, even when timeout is set to 0.

bin/php-language-server.php
felixfbecker: I don't get the second part of your sentence, why does it have to stop listening? I though it would then solely communicate through the socket?

MadHed: Only the listen socket is closed, the established connection keeps going. It just won't receive any additional connections after the first one.

felixfbecker: Ah okay so the server should only accept a single connection

felixfbecker:
@MadHed are you working on more tests? Would like to base more implementations on your refactoring :)

MadHed:
@felixfbecker I am :) just couldn't work on it the last few days. I hope to push a ready version later today.

src/ProtocolStreamReader.php
```diff
@@ -54,6 +54,9 @@ public function __construct($input)
                             $this->parsingMode = ParsingMode::HEADERS;
                             $this->headers = [];
                             $this->buffer = '';
+
+                            // after reading a full message, leave to allow different tasks to run
+                            return;
```
MadHed: We need some way to exit the while loop. (To serve setTimeouts, etc.) Without the return statement, the loop would only exit on EOF.

src/SymbolFinder.php
```diff
@@ -35,31 +36,65 @@ class SymbolFinder extends NodeVisitorAbstract
...
+
+        // If we enter a named node, push its name onto name stack.
+        // Else push the current name onto stack.
+        if (!empty($node->name) && (is_string($node->name) || method_exists($node->name, '__toString')) && !empty((string)$node->name)) {
+            if (empty($containerName)) {
+                array_push($this->nameStack, (string)$node->name);
+            }
+            else {
+                if ($node instanceof Node\Stmt\ClassMethod) {
+                    array_push($this->nameStack, $containerName . '::' . (string)$node->name);
```
MadHed: I don't know... probably, yes. I haven't got much experience with the PHPParser lib.

src/PhpDocument.php
```diff
@@ -0,0 +1,136 @@
...
+     * @return \PhpParser\Node[]
+     */
+    public function parse()
+    {
+        $stmts = null;
+        try {
+            $stmts = $this->parser->parse($this->content);
+        }
+        catch(Error $e) {
+            // Parser still throws errors. e.g for unterminated comments
```
MadHed: yup

bin/php-language-server.php
```diff
@@ -10,6 +12,21 @@
     }
 }
 
-$server = new LanguageServer(new ProtocolStreamReader(STDIN), new ProtocolStreamWriter(STDOUT));
+if (count($argv) >= 3 && $argv[1] == '--tcp') {
```
MadHed: Ok. There's also still the JS side that needs to be updated to actually use this connection type.

src/ProtocolStreamReader.php
felixfbecker: Hm - is there no way to only read until the buffer is empty? I thought you said when using a socket it would work non-blocking, so it would return an empty string as soon as the buffer is empty?

src/ProtocolStreamReader.php
felixfbecker: Sure, go ahead with updating

bin/php-language-server.php
MadHed: I will open a PR there

felixfbecker: If you are not familiar with Node / TypeScript, I can also do this :)

MadHed: That would be nice. I can open an issue and post what I have got so far. It kind of works but there are probably better ways to do it. ;)

MadHed:
Wrote tests for most of the new functionality. The current state of the windows blocking issue is as follows:

On most operating systems the communication between IDE and server can be handled via normal blocking stdio.
sabre/loop will issue a stream_select to check for incoming data with a timeout that is calculated based on the pending setTimeout queue.
As long as the StreamReader returns control to the event loop after a message has been read the loop will continue serving setTimeouts and incoming messages.

On windows stdio works differently and it is not really possible (at least with PHP) to check whether there is data available to read.
stream_select will always block until data is available regardless of the timeout value.
This causes the server to only process a single setTimeout callback after a message has arrived.

Sockets behave like normal streams on Windows/PHP and can be used instead. For a command line argument has been added to connect via a supplied tcp connection instead of stdio.
The extension.js will open a listen socket on Windows and start the server with the details to connect to it.

src/ProtocolStreamReader.php
MadHed: I think I wrote a lot of confusing stuff... :) please disregard everything I said earlier and read my comment about the current state of the event loop.

felixfbecker:
Looks pretty good. I still think though that returning early from the stream handler is correct. As far as I understand we now have a solution that allows non-blocking stream reads on all platforms. Therefor `fgetc` will return `""` when the buffer is empty, right? Therefor we can read the buffer till it's empty, and then the callback will exit and other events will get executed. Then when there is data again in the buffer, it will call the callback again. Or am I understanding this wrong? Maybe your PR fixed the behavior?

MadHed:
Yeah you are right. Committing the change.

felixfbecker:
Tested this an looks good to me! There was one case though were I met this error here:
```
Notice: fwrite(): send of 8192 bytes failed with errno=10035 No non-blocking socket task could not be executed.
 in C:\Users\felix\git\OpenSource\php-language-server\src\ProtocolStreamWriter.php on line 28
Call Stack:
    0.2230     370016   1. {main}() C:\Users\felix\git\OpenSource\php-language-server\bin\php-language-server.php:0
    0.2869    4141816   2. Sabre\Event\Loop\run() C:\Users\felix\git\OpenSource\php-language-server\bin\php-language-server.php:34
    0.2869    4141816   3. Sabre\Event\Loop\Loop->run() C:\Users\felix\git\OpenSource\php-language-server\vendor\sabre\event\lib\Loop\functions.php:122
  129.3512   19208048   4. Sabre\Event\Loop\Loop->tick() C:\Users\felix\git\OpenSource\php-language-server\vendor\sabre\event\lib\Loop\Loop.php:194
  129.3512   19208048   5. Sabre\Event\Loop\Loop->runStreams() C:\Users\felix\git\OpenSource\php-language-server\vendor\sabre\event\lib\Loop\Loop.php:233
  134.4875   19208552   6. LanguageServer\ProtocolStreamReader->LanguageServer\{closure}() C:\Users\felix\git\OpenSource\php-language-server\vendor\sabre\event\lib\Loop\Loop.php:311
  134.4893   19210672   7. LanguageServer\LanguageServer->LanguageServer\{closure}() C:\Users\felix\git\OpenSource\php-language-server\src\ProtocolStreamReader.php:52
  134.6761   19477648   8. LanguageServer\ProtocolStreamWriter->write() C:\Users\felix\git\OpenSource\php-language-server\src\LanguageServer.php:69
  134.6949   33229832   9. fwrite() C:\Users\felix\git\OpenSource\php-language-server\src\ProtocolStreamWriter.php:28
```
Any idea how we can prevent this?

felixfbecker:
I think it is a temporary solution and only a notice. We should silence it with an `@` (and handle real errors ourself)

MadHed:
> Tested this an looks good to me! There was one case though were I met this error here: [...]

Ugh... didn't even think about this.
Since the connection is now nonblocking for input and output we need to buffer the outgoing messages.

MadHed:
Can you reproduce the StreamWriter bug? It should be fixed now.

src/ProtocolStreamWriter.php
```diff
@@ -25,6 +25,16 @@ public function __construct($output)
      */
     public function write(Message $msg)
     {
-        fwrite($this->output, (string)$msg);
+        $data = (string)$msg;
+        $msgSize = strlen($data);
+        $totalBytesWritten = 0;
+
+        while ($totalBytesWritten < $msgSize) {
+            $bytesWritten = fwrite($this->output, substr($data, $totalBytesWritten));
```
felixfbecker: We should suppress the notice triggered with `@`

src/ProtocolStreamWriter.php
```diff
@@ -25,6 +25,16 @@ public function __construct($output)
...
+        $msgSize = strlen($data);
+        $totalBytesWritten = 0;
+
+        while ($totalBytesWritten < $msgSize) {
+            $bytesWritten = @fwrite($this->output, substr($data, $totalBytesWritten));
+            if ($bytesWritten === false) {
+                throw new Error('Could not write message.');
+            }
+            $totalBytesWritten += $bytesWritten;
+        }
```
felixfbecker: Could you explain a bit more what you are doing here? As I understand we block until everything is written out, is this intended? Should we use a setTimeout or nextTick here?

Also, `Error` is reserved for PHP internal errors, should be `Exception`

MadHed: If we write more data than the output buffer can hold we get partial writes, so we loop until all data is written.
This didn't happen before because in blocking mode this loop is handled internally.

We need to do this synchronously because otherwise we could end up with partial messages intermingled with each other.

Will change to `Exception`

felixfbecker:
@MadHed Do you feel this is ready to merge?

I also did some final benchmark. Indexing this project on a Surface Pro 3 (i5-4300U):

- 1299 files, ~13000 physical lines of code
- 41 seconds, 64 MiB memory
- 0.032s / file, 0.0031s / PLOC
- 0.0493 MiB / file, 0.0049 MiB / PLOC

MadHed:
@felixfbecker I am not aware of any remaining bugs, so if you want, go ahead.

I am also using this version productively on an internal PHP Project with 2613 files and had no crashes or anything. 

Even though some features need improvement I think this will serve as a great starting point for future development.

Ahmad-Alhourani:
@MadHed 
plz I want using this feature by jsonrpc, how  I can send the request and mydir through it.

felixfbecker:
@Ahmad-Alhourani that question would be better suited for StackOverflow as a general question about LSP / JSON RPC.

PR #32, opened 2016-09-20
Use PHP-CS-Fixer as a formatter

mniewrzal:


fixtures/format.php
```diff
@@ -9,12 +9,16 @@ class TestClass
...
         $testVariable = 123;
-        
-                if (  empty($testParameter)){
-                    echo 'Empty';
+
+        if (  empty($testParameter))    {
+            echo 'Empty';
         }
     }
 }
```
felixfbecker: Why the changes in this file?

mniewrzal: CS-Fixer works in different way than Pretty Printer. PP is converting ast into source code according to some rules. CS-Fixer search for invalid part of code and fix just that part. Looks that not all parts of source code are covered by cs-fixer and for example "namespace       TestNamespace;" wont be converted to "namespace TestNamespace;". I still need to check if its by design or its a bug.

mniewrzal:
Code Sniffer formatter looks to be better choice in my opinion.

PR #35, opened 2016-09-27
Use PHP_CodeSniffer as a formatter

mniewrzal:


src/Server/Formatter.php
```diff
@@ -0,0 +1,63 @@
+<?php
+namespace LanguageServer\Server;
+
+use LanguageServer\Protocol\TextEdit;
+use LanguageServer\Protocol\Range;
+use LanguageServer\Protocol\Position;
```
felixfbecker: Please use a group use statement here

src/Server/Formatter.php
```diff
@@ -0,0 +1,63 @@
...
+    /**
+     * @param string $content
+     * @param string $uri
+     *
+     * @return TextEdit[]
+     */
+    public function format(string $content, string $uri)
+    {
+        // remove 'file://' prefix
+        $uri = substr($uri, 7);
```
felixfbecker: I don't think this is the right way to do it - an input of `file:///c:/users/felix/test` would result in `/c:/users/felix/test` which is not a valid path. #31 adds a `path2uri` function, I would propose to add the reverse `uri2path` function (with some tests) https://github.com/MadHed/php-language-server/blob/7b0b9b890cabd3f0c2e1d7f10314ed61f53df110/src/utils.php#L23-L33

src/Server/Formatter.php
```diff
@@ -0,0 +1,63 @@
...
+            $default = $currentDir . DIRECTORY_SEPARATOR . 'phpcs.xml.dist';
+            if (is_file($default) === true) {
+                return array($default);
+            }
+            
+            $lastDir = $currentDir;
+            $currentDir = dirname($currentDir);
+        } while ($currentDir !== '.' && $currentDir !== $lastDir);
+        
+        $standard = \PHP_CodeSniffer::getConfigData('default_standard');
```
felixfbecker:
```php
$standard = \PHP_CodeSniffer::getConfigData('default_standard') ?? 'PSR2'
```

src/Server/Formatter.php
mniewrzal: For now I added new method to Formatter to avoid conflicts with #31. Later I can move it to utils.php

src/Server/Formatter.php
mniewrzal: Done

src/Server/Formatter.php
mniewrzal: Done

kaloyan-raev:
I noticed that a big mess happens when formatting the SymbolFinder.php file - in this part: https://github.com/felixfbecker/php-language-server/blob/master/src/SymbolFinder.php#L54-L63

mniewrzal:
Code Sniffer is failing to format SymbolFinder.php. I added simple error handling. Tomorrow I will report this problem to PHPCS.

src/Server/Formatter.php
```diff
@@ -11,15 +12,19 @@ class Formatter
...
     public function format(string $content, string $uri)
     {
         $path = $this->uriToPath($uri);
         $cs = new \PHP_CodeSniffer();
         $cs->initStandard($this->findConfiguration($path));
         $file = $cs->processFile(null, $content);
-        $file->fixer->fixFile();
+        $fixed = $file->fixer->fixFile();
+        if (!$fixed) {
+            throw new ResponseError("Unable to format file", ErrorCode::INTERNAL_ERROR);
```
felixfbecker: Please use single quotes

mniewrzal:
I reported formatting issue: https://github.com/squizlabs/PHP_CodeSniffer/issues/1180

felixfbecker:
Will merge #31 first, will probably some refactoring for you to rebase

mniewrzal:
Ok, I'm aware of that ;)

felixfbecker:
Just merged #31

src/Protocol/TextEdit.php
```diff
@@ -22,4 +22,10 @@ class TextEdit
      * @var string
      */
     public $newText;
+
+    public function __construct(Range $range = null, string $newText)
```
felixfbecker: Non-optional parameter cannot follow an optional param. Both should be optional to allow constructing a instance and then setting the properties manually

tests/Utils/FileUriTest.php
```diff
@@ -33,4 +33,25 @@ public function testBackslashesAreTransformed()
...
+    
+    public function testUriToPathUnix()
+    {
+        $uri = "file:///home/foo/bar/Test.php";
+        $this->assertEquals("/home/foo/bar/Test.php", \LanguageServer\uriToPath($uri));
+    
+        $uri = "/home/foo/bar/Test.php";
+        $this->assertEquals("/home/foo/bar/Test.php", \LanguageServer\uriToPath($uri));
+    
+        $uri = "/home/foo space/bar/Test.php";
```
felixfbecker: Not a valid URI, space needs to be `%20`

tests/Utils/FileUriTest.php
```diff
@@ -33,4 +33,25 @@ public function testBackslashesAreTransformed()
...
+        $this->assertEquals("/home/foo space/bar/Test.php", \LanguageServer\uriToPath($uri));
+    }
+    
+    public function testUriToPathWindows()
+    {  
+        $uri = "file:///c:/home/foo/bar/Test.php";
+        $this->assertEquals("c:\\home\\foo\\bar\\Test.php", \LanguageServer\uriToPath($uri));
+    
+        $uri = "c:\\home\\foo\\bar\\Test.php";
```
felixfbecker: That's not a valid URI so I don't think we need to test against it

src/utils.php
```diff
@@ -32,3 +32,19 @@ function pathToUri(string $filepath): string {
...
+ * @return string
+ */
+function uriToPath(string $uri)
+{
+    $position = strpos($uri, '://');
+    if (! $position) {
+        return $uri;
+    }
+    $path = substr($uri, $position + 3);
+    return strpos($path, ':') ? str_replace('/', "\\", substr($path, 1)) : $path;
```
felixfbecker: Not that happy with this implementation.
- If possible I would like to use [`parse_url`](http://php.net/manual/en/function.parse-url.php)
- Needs to handle %-encoding
- Needs to replace `/` with `DIRECTORY_SEPERATOR`

tests/Server/TextDocumentTest.php
```diff
@@ -218,6 +218,16 @@ public function testFormatting()
...
     }
+    
+    public function testFormattingInvalidUri()
+    {
+        $client =  new LanguageClient(new MockProtocolStream());
+        $project = new Project($client);
+        $textDocument = new Server\TextDocument($project, $client);
+        
+        $result = $textDocument->formatting(new TextDocumentIdentifier('whatever'), new FormattingOptions());
+        $this->assertEquals([], json_decode(json_encode($result), true));
```
felixfbecker: Imo an invalid URI should throw an error instead

src/Formatter.php
```diff
@@ -0,0 +1,65 @@
...
+     * @param string $uri            
+     * @return string[]
+     */
+    private function findConfiguration(string $uri)
+    {
+        $currentDir = dirname($uri);
+        do {
+            $default = $currentDir . DIRECTORY_SEPARATOR . 'phpcs.xml';
+            if (is_file($default) === true) {
+                return array($default);
```
felixfbecker: Use short array syntax

tests/Server/TextDocumentTest.php
mniewrzal: Actual implementation will create `PhpDocument` for invalid/non-existing uri (`Project::getDocument`). I can try to reorganize this with this change but I think it should be done as separate issue.

src/utils.php
```diff
@@ -41,10 +41,6 @@ function pathToUri(string $filepath): string {
...
 function uriToPath(string $uri)
 {
-    $position = strpos($uri, '://');
-    if (! $position) {
-        return $uri;
-    }
-    $path = substr($uri, $position + 3);
-    return strpos($path, ':') ? str_replace('/', "\\", substr($path, 1)) : $path;
+    $path = urldecode(parse_url($uri)['path']);
+    return strpos($path, ':') ? str_replace(\DIRECTORY_SEPARATOR, '\\', substr($path, 1)) : $path;
```
felixfbecker: - `strpos` may return `0` or `false`, should check explicitly for `false` here
- It's `str_replace($search, $replace, $subject)`, we need to search for `'/'` and replace it with `DIRECTORY_SEPERATOR`

tests/Utils/FileUriTest.php
```diff
@@ -34,24 +34,24 @@ public function testBackslashesAreTransformed()
...
-    
-        $uri = "c:\\home\\foo\\bar\\Test.php";
-        $this->assertEquals("c:\\home\\foo\\bar\\Test.php", \LanguageServer\uriToPath($uri));
+        $uri = 'file:///d/e/f';
+        $this->assertEquals('/d/e/f', \LanguageServer\uriToPath($uri));
+        
+        $uri = 'file:///a/b/c/test.txt';
+        $this->assertEquals('/a/b/c/test.txt', \LanguageServer\uriToPath($uri));
+        
+        $uri = 'file:///c%3A/foo/bar.baz';
```
felixfbecker: The colon after the drive letter is typically not encoded afaik
https://en.wikipedia.org/wiki/File_URI_scheme#Windows_2

src/Formatter.php
```diff
@@ -0,0 +1,65 @@
...
+        $fixed = $file->fixer->fixFile();
+        if (!$fixed && $file->getErrorCount() > 0) {
+            throw new ResponseError('Unable to format file', ErrorCode::INTERNAL_ERROR);
+        }
+        
+        $new = $file->fixer->getContents();
+        if ($content === $new) {
+            return [];
+        }
+        return [new TextEdit(new Range(new Position(0, 0), new Position(PHP_INT_MAX, PHP_INT_MAX)), $new)];
```
kaloyan-raev: Eclipse Che requires an accurate calculation of the end position. Although VSCode can handle this well, the editor in Che just breaks if any negative values or big values beyond the document end are provided.

mniewrzal: Ok, I will try to calculate end position.

kaloyan-raev: Thanks! This now works very well in Che.

tests/Utils/FileUriTest.php
mniewrzal: Changed to simple colon.

kaloyan-raev:
@felixfbecker is there anything else pending (except resolving conflicts) for this PR to be merged?

src/Formatter.php
```diff
@@ -0,0 +1,75 @@
...
+     * @param string $uri            
+     *
+     * @return \LanguageServer\Protocol\TextEdit[]
+     * @throws \AdvancedJsonRpc\ResponseError
+     */
+    public function format(string $content, string $uri)
+    {
+        $path = uriToPath($uri);
+        $cs = new \PHP_CodeSniffer();
+        $cs->initStandard($this->findConfiguration($path));
```
felixfbecker: Will this also work in environments where
- URIs are _not_ file URIs
- there is no config file in the project
?

src/utils.php
```diff
@@ -32,3 +32,15 @@ function pathToUri(string $filepath): string {
...
+
+/**
+ * Transforms URI into file path
+ * 
+ * @param string $uri
+ * @return string
+ */
+function uriToPath(string $uri)
+{
+    $filepath = urldecode(parse_url($uri)['path']);
```
felixfbecker: A URI can have any protocol, but only file URIs can be mapped to a path. We need to throw an `InvalidArgumentException` if the URI is not a `file://` URI and also handle/test that case in the implementation.

src/utils.php
```diff
@@ -32,3 +32,15 @@ function pathToUri(string $filepath): string {
...
+    return strpos($filepath, ':') === false ? $filepath : str_replace('/', '\\', $filepath);
```
felixfbecker: This is a little brittle. There are also implementations that use `|` instead of the colon on windows. You don't need to test against that, but it would be better to simply do `str_replace('/', DIRECTORY_SEPERATOR, $filepath)`

https://en.wikipedia.org/wiki/File_URI_scheme#Windows

src/Formatter.php
```diff
@@ -0,0 +1,75 @@
...
+    }
+
+    /**
+     * @param string $content
+     * @return \LanguageServer\Protocol\Position
+     */
+    private function calculateEndPosition(string $content): Position
+    {
+        $lines = explode("\n", $content);
+        return new Position(sizeof($lines) - 1, strlen(end($lines)));
```
felixfbecker: Please use `count`, `sizeof` is just an alias and PHP has so many aliases nobody can remember them all 😄

src/Formatter.php
```diff
@@ -0,0 +1,75 @@
...
+    }
+    
+    /**
+     *
+     * @param string $uri            
+     * @return string[]
+     */
+    private function findConfiguration(string $uri)
```
felixfbecker: The parameter is named `$uri`, but you call it with a path, not a URI. Some description for the function and params would also be nice

src/Formatter.php
```diff
@@ -0,0 +1,75 @@
...
+    {
+        $currentDir = dirname($uri);
+        do {
+            $default = $currentDir . DIRECTORY_SEPARATOR . 'phpcs.xml';
+            if (is_file($default) === true) {
```
felixfbecker: Unneeded `=== true`

src/Formatter.php
```diff
@@ -0,0 +1,75 @@
...
+     */
+    public function format(string $content, string $uri)
+    {
+        $path = uriToPath($uri);
+        $cs = new \PHP_CodeSniffer();
+        $cs->initStandard($this->findConfiguration($path));
+        $file = $cs->processFile(null, $content);
+        $fixed = $file->fixer->fixFile();
+        if (!$fixed && $file->getErrorCount() > 0) {
+            throw new ResponseError('Unable to format file', ErrorCode::INTERNAL_ERROR);
```
felixfbecker: We don't need to throw a `ResponseError`, any Exception will be turned into an `ErrorResponse` with `ErrorCode::INTERNAL_ERROR` by the dispatcher automatically. No need to couple it to JSON RPC here

tests/FormatterTest.php
```diff
@@ -0,0 +1,32 @@
...
+    public function testFormatNoChange()
+    {
+        $formatter = new Formatter();
+        $expected = file_get_contents(__DIR__ . '/../fixtures/format_expected.php');
+        
+        $edits = $formatter->format($expected, 'whatever');
+        $this->assertTrue($edits == []);
+    }
+
+}
```
felixfbecker: Missing newline - I don't know which editor you use but I recommend you to look for an EditorConfig extension 😉

src/Formatter.php
```diff
@@ -0,0 +1,75 @@
+<?php
+declare(strict_types = 1);
+
+namespace LanguageServer;
+
+use LanguageServer\Protocol\{TextEdit, Range, Position, ErrorCode};
+use AdvancedJsonRpc\ResponseError;
+
+class Formatter
```
felixfbecker: Initially I was wondering why this needs to be a class that is instantiated when it has no properties, constructor or internal state. Everything could just be pure functions. But it's probably better to encapsulate the private functions. One possibility would be though to make them static and the class `abstract`. Opinions on this?

tests/Server/TextDocumentTest.php
felixfbecker: Fair enough.

tests/Server/TextDocumentTest.php
```diff
@@ -211,13 +211,23 @@ public function testFormatting()
...
```
felixfbecker: Don't need to `json_decode(json_encode())`

src/Formatter.php
```diff
@@ -0,0 +1,75 @@
...
+     * @param string $content            
+     * @param string $uri            
+     *
+     * @return \LanguageServer\Protocol\TextEdit[]
+     * @throws \AdvancedJsonRpc\ResponseError
+     */
+    public function format(string $content, string $uri)
+    {
+        $path = uriToPath($uri);
+        $cs = new \PHP_CodeSniffer();
```
felixfbecker: Please add a `use PHP_CodeSniffer` at the top. This declares the dependency on the class and avoids having to use absolute references

src/Formatter.php
mniewrzal: I added this class to prevent messing with PhpDocument in future. Abstract class and static methods sounds ok for me.

src/Formatter.php
mniewrzal: > Will this also work in environments where
> 
> URIs are not file URIs

For directory URI it wont work if URI = directory with config file but in this case it will load default configuration. Should we be worried about such cases? Do you notices URIs different than file?

> there is no config file in the project ?

If config file is not found PSR2 standard is loaded.

felixfbecker:
Fixed the style issues

src/utils.php
felixfbecker: Forget what I said. the directory separator should of course be determined by the URI, not the runtime OS.

felixfbecker:
I think this is good to go, just needs a rebase

src/Formatter.php
felixfbecker: What I meant was that we need to handle URIs that don't use the `file://` scheme, but something like `vfs://`.

PR #37, opened 2016-09-29
Add phpunit configuration file

mniewrzal:
This file will preconfigure phpunit options. No need to write 'vendor/bin/phpunit --bootstrap vendor/autoload.php tests' anymore.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,7 @@ install:
  - composer install
 
 script:
- - vendor/bin/phpunit --coverage-clover=coverage.xml --whitelist src --bootstrap vendor/autoload.php tests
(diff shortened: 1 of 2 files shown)


felixfbecker:
What does the `.xml.dist` mean? Sorry never used the XML config before

mniewrzal:
This file can have .xml or .xml.dist extension. PHPUnit first search for xml and if not found for xml.dist. Extension 'dist' stands for distribution. Such version contains basic parameters for all developers. If you want to customize parameters then you can create phpunit.xml with own version of this file and framework will use it. In this case someone can create phpunit.xml will additional configuration for 'coverage-html' for local usage.

PR #38, opened 2016-09-29
Initial implemention for code completion

mniewrzal:
This is initial draft for keywords and local variables completion. Provides:
- keywords from predefined list
- all variables from file
- simple mechanism to add new completion items

In future list of variables and keywords should be filtered according to completion context (global/class/method) but it shouldn't be a big problem. For now this PR is to show conception. If it will be approved then I will prepare version with more phpDocs/tests/etc. Few things in code also should be reorganized.


src/Protocol/Range.php
```diff
@@ -26,4 +26,14 @@ public function __construct(Position $start = null, Position $end = null)
         $this->start = $start;
         $this->end = $end;
     }
+
+    public function match(Position $position)
+    {
+        return $this->start->compare($position) <= 0 && $this->end->compare($position) >= 0;
+    }
+    
+    public function toString()
```
felixfbecker: Should be `__toString()`

src/Server/Completion/Strategies/KeywordsStrategy.php
```diff
@@ -0,0 +1,116 @@
...
+            new KeywordData("throw"),
+            new KeywordData("trait"),
+            new KeywordData("true"),
+            new KeywordData("try"),
+            new KeywordData("unset"),
+            new KeywordData("use"),
+            new KeywordData("var"),
+            new KeywordData("while"),
+            new KeywordData("xor"),
+            new KeywordData("yield")
```
felixfbecker: Kind of dislike this class. Seems to me like all these are just constants, so they should be constants imo.
Also, where do these come from? The perfect solution would be to parse them somewhere.

src/Server/Completion/TokenContainer.php
```diff
@@ -0,0 +1,80 @@
...
+
+    const T_SINGLE_CHAR = 10000;
+
+    /**
+     *
+     * @var \LanguageServer\Server\Completion\Token[]
+     */
+    private $tokens;
+
+    public function __construct(array $tokens)
```
felixfbecker: Is it needed to work with the raw PHP tokens? Could you comment on the limitations of PHPParser in that regard? The error recovery should be pretty good by now: https://github.com/nikic/PHP-Parser/issues/262
I would like to work with the AST as much as possible (and if it is not capable, consider contributions).

src/Server/TextDocument.php
```diff
@@ -148,4 +167,14 @@ public function formatting(TextDocumentIdentifier $textDocument, FormattingOptio
         return [$edit];
     }
     
+    public function completion(TextDocumentIdentifier $textDocument, Position $position)
+    {
+        $completionReporter = new CompletionReporter();
+        $tokens = token_get_all($this->documents[$textDocument->uri]);
```
felixfbecker: See above comment, this will essentially re-parse the whole document, even though PHPParser already did this. If PHPParser is really not capable of parsing nodes up until the cursor, we should take as many nodes as it can provide and then only use `token_get_all()` for the substring between the last node and the cursor, if `token_get_all` will even work for this. Might even have to use regexps for those last chars in front of the cursor.

src/Protocol/Range.php
mniewrzal: Ok, I will update PR with this and other suggestions.

src/Server/Completion/Strategies/KeywordsStrategy.php
mniewrzal: This is more like example of usage then final version. I agree that it can be transformed in [][]. List of keywords can be found here: http://php.net/manual/en/reserved.keywords.php. It would be nice to have parsable keywords list, for now hardcoded list will be good enough.

src/Server/Completion/TokenContainer.php
mniewrzal: This simple example will return null instead of ast:
```
<?php
function abc(){
}
function test(){
    $
```
From my personal experience I'm concerned that there are too much cases to cover to have very good parser error recovery. I will explain more about tokens in next comment.

felixfbecker: What about
```php
function abc(){
}
function test(){
    $
}
```
?
VS Code automatically inserts closing brackets

src/Server/TextDocument.php
mniewrzal: This part was written just to show results, I knew I will need to rebase this PR after #31 merge. 

List of tokens is always available after Parser::parse(), Lexer::getTokens() contains such array and it can be used without executing token_get_all() once again. Even without that dividing into tokens is very fast :) Main problem with pure tokens array is that its a little bit inconsistent and it needs to be converted into something more useful for line:column processing. In this case conversion changes all arrays elements into the same object type with document range. 

Mixing ast with token_get_all is in theory possible because token_get_all can process part of code ('<?php' tag needs to be added) but I'm not sure if its best solution. It can be tricky in my opinion.

Token approach looks right now to be good solution in my opinion but its just my version and maybe someone will came with better implementation :)

src/Server/Completion/TokenContainer.php
kaloyan-raev: Another use case is typing just:

```php
<?php
abstract
```

In this case the PHP parser returns an empty AST tree.

The only way to find all situations when AST tree is not enough seems to be by trial and error. So, it looks reasonable to rely on tokens than on AST.

mniewrzal:
I changed a little bit previous implementation. Now it doesn't use tokens just regexp in some cases. With this patch CA provides:

- local variables
- keywords
- class members (no inheritance)
- global elements (functions, types)

felixfbecker:
@mniewrzal Really sorry this PR has been stale for so long. It includes a lot of changes for one of the major and most complicated features, so I would like to think about architecture design before merging anything, and I'm currently focused on #54, which completion will also benefit from. You are introducing a lot of new classes like strategies, resolvers and I need to tinker a bit myself to see if that added complexity makes sense.

But I would be interested in your experience with providing the completion: My idea was that we use the AST and get the last nodes before the cursor. I haven't played with it a lot, but did you try this? Because then we can let generic functions handle resolving a property fetch, an array dimension fetch or a method/function call recursively to a _type_. We then would only need a regexp or tokens for the characters between the last node and the cursor.

felixfbecker:
Completion support landed in v4.0.0, thanks for your efforts!

PR #43, opened 2016-09-30
Fixed issue #42 and added regression test

### LINKED ISSUE #42: Error in SymbolFinder when processing variable variables
When processing variable variables, i.e `$$foo` SymbolFinder breaks because these nodes do not have a `name` property that is castable to `string`.

MadHed:
This should fix #42


felixfbecker:
Just a tip, your git email is not known by GitHub

MadHed:
Now it is

felixfbecker:
The git force is strong with this one 👆

PR #49, opened 2016-10-08
Go to definition support

### LINKED ISSUE #19: Goto definition
https://github.com/Microsoft/language-server-protocol/blob/master/protocol.md#goto-definition-request

felixfbecker:
Closes #19

animation

Classes, interfaces, traits, methods, properties and constants get indexed by their fully qualified name.
Examples for FQNs:
 - testFunction()
 - TestNamespace\TestClass
 - TestNamespace\TestClass::TEST_CONSTANT
 - TestNamespace\TestClass::staticTestProperty
 - TestNamespace\TestClass::testProperty
 - TestNamespace\TestClass::staticTestMethod()
 - TestNamespace\TestClass::testMethod()

Variables are resolved at request.

So far only Gotodef for classes is tested, other tests will follow.

What does not work yet:
```php
$x = new Whatever();
$y = $x;
$y->myMethod(); // cannot get definition for myMethod

$x = new Whatever();
$x->property->myMethod(); // cannot get definition for myMethod

$this->myMethod(); // cannot get definition for myMethod, $this is not resolved to the class

function test_function(TestClass $x)  // cannot get definition for TestClass
{
}
```

Need to add some recursive logic there.


felixfbecker:
Anyone have an idea what kind of sorcery this is?
```php
$document = $this->project->getDefinitionDocument($name);
```
```
Error: Cannot access private property LanguageServer\PhpDocument::$project
C:\Users\felix\git\OpenSource\php-language-server\src\PhpDocument.php:345
```
How can this error ever happen when reading from `$this`?!

src/NodeVisitors/NodeAtPositionFinder.php
```diff
@@ -0,0 +1,47 @@
...
+    public function __construct(Position $position)
+    {
+        $this->position = $position;
+    }
+
+    public function leaveNode(Node $node)
+    {
+        if (
+            !isset($this->node)
+            && $node->getAttribute('startLine') <= $this->position->line + 1
```
MadHed: looks good to me

MadHed:
That's a lot of changes... can you give a summary when you are finished?

felixfbecker:
@MadHed All the latest commits were pretty much bug fixes. I tried to make the commits easy to review one-by-one (actually spend quite some time rebasing).

Test refactoring etc. should be self-explanatory.

4786fe173ce690eb176f847e89b0456df7d44596, befb07d5c9decb1a822ea7d5eb5857fa211bf294: This adds new attributes to each Node that allow you to get the parent node, siblings and the document of the node. Inspired by  
  https://developer.mozilla.org/en-US/docs/Web/API/Node/parentNode  
  https://developer.mozilla.org/en-US/docs/Web/API/Node/nextSibling  
  https://developer.mozilla.org/en-US/docs/Web/API/Node/previousSibling  
  https://developer.mozilla.org/en-US/docs/Web/API/Node/ownerDocument  

Implementation of definition request (9f2d3a4fbc4993529bea493b02ef91b13b18c8a0):
Upon parsing, a new NodeVisitor `DefinitionCollector` traverses the AST(s) and collects all declarations (except variables). The declaration nodes are stored in a map by their (unique) FQN. Methods/properties are separated from their classes with `::`, methods and functions are appended `()`.
These definitions are then stored in the `PhpDocument` object under `$definitions`.
At the same time, the FQNs are also registered at the `Project` object to point to this `PhpDocument` object.

To get the definition, we first have to find the node under the cursor. This is done by the `NodeAtPositionFinder` (2fd3b58f74af245365939855edb7169dd904dff8), which traverses the AST depth-first and returns the first node that contains the cursor position. This requires to keep the AST in memory, which is now saved under `$stmts` on the `PhpDocument` object.

Once we have the node we want to get the definition for, we build the FQN from it.
This happens in [`PhpDocument::getDefinitionByNode()`](https://github.com/felixfbecker/php-language-server/pull/49/files#diff-eb1eb31b9e300fdb6107f9b506d72f36R283). We basically have to go through all possible nodes one could ask the definition for (property fetch, function call, method call, ...) and try to resolve the FQN of the symbol.
This is pretty easy for classes or static method calls for example.
Method calls, property fetches and variables are more complicated. We first have to find an assignment or parameter node that this variable comes from, and then hopefully can infer it from there.
This is done in [`PhpDocument::getVariableDefinition()`](https://github.com/felixfbecker/php-language-server/pull/49/files#diff-eb1eb31b9e300fdb6107f9b506d72f36R391).
The algorithm is as follows: From the variable node, check if any previous siblings are an assignment node to that variable. If not, go one level up. Repeat this until a function/method node is met, and finally check parameters.
The types are resolved very basic atm, it only works if the expression is a `New` node (instantiation) or if the parameter has a type hint. It breaks when the variable is assigned to another variable or the return value of a function. DocBlocks are also not taken into account at all. I want to save this for later implementation, will become more important for `provideCompletion`. The biggest use cases for `definition` are implemented in this PR.

Finally, I removed the `SymbolFinder` and instead use the `$definitions` array to build the `SymbolInformation` array.

felixfbecker:
Squashed some commits to easy reviewing.

MadHed:
Ok let me check this out.

src/PhpDocument.php
```diff
@@ -201,4 +246,163 @@ public function getNodeAtPosition(Position $position)
...
+                return null;
+            }
+            $name = (string)$node->class;
+        } else if ($node instanceof Node\Expr\MethodCall || $node instanceof Node\Expr\PropertyFetch) {
+            if ($node->name instanceof Node\Expr) {
+                // Cannot get definition of dynamic calls
+                return null;
+            }
+            // Need to resolve variable to a class
+            $varDef = $this->getVariableDefinition($node->var);
```
MadHed: This crashes with fatal when trying to goto definition of `write` in `$this->protocolWriter->write`.

`Fatal error:  Uncaught TypeError: Argument 1 passed to LanguageServer\PhpDocument::getVariableDefinition() must be an instance of PhpParser\Node\Expr\Variable, instance of PhpParser\Node\Expr\PropertyFetch given, called in E:\Projekte\php-language-server\src\PhpDocument.php on line 315 and defined in E:\Projekte\php-language-server\src\PhpDocument.php:391`

src/PhpDocument.php
```diff
@@ -70,9 +89,36 @@ public function __construct(string $uri, Project $project, LanguageClient $clien
...
+            Node\Stmt\Trait_::class           => SymbolKind::CLASS_,
+            Node\Stmt\Interface_::class       => SymbolKind::INTERFACE,
+            Node\Stmt\Namespace_::class       => SymbolKind::NAMESPACE,
+            Node\Stmt\Function_::class        => SymbolKind::FUNCTION,
+            Node\Stmt\ClassMethod::class      => SymbolKind::METHOD,
+            Node\Stmt\PropertyProperty::class => SymbolKind::PROPERTY,
+            Node\Const_::class                => SymbolKind::CONSTANT
+        ];
+        $symbols = [];
+        foreach ($this->definitions as $fqn => $node) {
```
MadHed: `Warning: Invalid argument supplied for foreach` when invoking `workspace/symbol` search

src/NodeVisitors/DefinitionCollector.php
```diff
@@ -0,0 +1,72 @@
...
+     *
+     * @var Node[]
+     */
+    public $definitions = [];
+
+    public function enterNode(Node $node)
+    {
+        if ($node instanceof Node\Stmt\ClassLike && isset($node->name)) {
+            // Class, interface or trait declaration
+            $this->definitions[(string)$node->namespacedName] = $node;
```
felixfbecker: I read `isset($node->name)` as _does the class have a name_ e.g. _is the class an anonymous class_ which is why I wrote it like this

MadHed:
Works for the trivial cases, but feels a little brittle all around... what do you think?
We definitely need to do a lot of testing to prevent crashing.

Also some time later we should switch to a custom AST representation like @mniewrzal proposed in https://github.com/felixfbecker/php-language-server/pull/31#issuecomment-247930695 this would probably decrease the memory footprint a lot and also tidy up the code quite a bit. I was a sometimes bit lost while reviewing 😁 

Also I just noticed when there are multiple assignment to a variable and you go to its definition, it just jumps up one assignment instead of going to the first one.

src/PhpDocument.php
felixfbecker: Can you reproduce? Maybe this is only when parsing hasn't finished yet?

src/NodeVisitors/DefinitionCollector.php
MadHed: Fair enough, it just looked like a bug at first because you test for one attribute but access another.

felixfbecker:
> Works for the trivial cases, but feels a little brittle all around... what do you think?

Yeah well the type resolving is still not very good. That will play a bigger role for provideCompletion. I want to get basic support for the major requests asap

> We definitely need to do a lot of testing to prevent crashing.

For sure. `DefinitionTest` is already the biggest test class

> Also some time later we should switch to a custom AST representation like @mniewrzal proposed in #31 (comment) this would probably decrease the memory footprint a lot and also tidy up the code quite a bit. I was a sometimes bit lost while reviewing 😁 

Still not 100% sold on this. The AST nodes are very very powerful to work with. Recreating them all seems a bit pointless to me. But what we can do now is not keep the file _content_ in memory anymore, because we only need it at parse time for errors and use the AST afterwards. That alone will reduce memory usage (the AST doesn't hold any content, only structure), but as I said already in other places I'm not concerned at the moment with RAM usage, rather with features and response times.

> Also I just noticed when there are multiple assignment to a variable and you go to its definition, it just jumps up one assignment instead of going to the first one.

Yeah, noticed that later on too. The algorithms picks the closest definition. Instead, it should go all the way up to a function node and then down again to find the declaration. But that is much more complicated because you need to make sure you are going the right path.

src/PhpDocument.php
MadHed: Can reproduce, let the indexer run all the way through. Every time(!) I open workspace symbol search the warning appears. The search works, though.

src/PhpDocument.php
MadHed: Yep, fixed

felixfbecker:
> Furthermore I was not able to goto definition of parameter type hints. Is this expected?

Fixed

src/PhpDocument.php
felixfbecker: I can reproduce too. But it is weird because `parse()` sets it to an array and `getSymbols()` never runs before `parse()`.

MadHed: I just found it. It comes from `invalid_file.php`. It's never successfully parsed.

MadHed: Fixed it

src/PhpDocument.php
```diff
@@ -53,7 +53,7 @@ class PhpDocument
      *
      * @var Node[]
      */
-    private $stmts;
+    private $stmts = [];
```
felixfbecker: I don't think this is good - there is a semantic difference between "a document has no statements" and "we couldn't parse the statements"

felixfbecker:
Please avoid merge commits

PR #52, opened 2016-10-09
References

felixfbecker:
image

Adds support for `textDocument/references`. At index time in addition to definitions, also a map of FQN to array of referencing document URIs is kept.

Just like definition, it works very well for class-likes, functions and constants, but not so great for methods and properties where it is hard to infer the class. Variables and parameters also work surprisingly well.

# ToDo
- [x] Tests for other symbols than classes

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -24,7 +24,7 @@
     "bin": ["bin/php-language-server.php"],
     "require": {
         "php": ">=7.0",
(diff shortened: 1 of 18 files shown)


felixfbecker:
Changed reference to only include class name
image

felixfbecker:
Contains only tests for class-likes, not functions or variables

Issue #47, opened 2016-10-07
Cache indexing result


MadHed:
Currently the project index is rebuilt from scratch every time the workspace is opened.
This can take a long time during which features might not work properly.

We need a file system cache for the results of the indexing process.

felixfbecker:
Definitely, but I would prioritize features for now.

MadHed:
Sure, I just wanted to have this as an issue so discussion about the implementation can take place.

PR #56, opened 2016-10-10
Add stubs for PHP built-ins

felixfbecker:
This is just a proposal on how the Jetbrains PHPStorm stubs could be added as a dependency.
We would then also need to include these in the indexing.

Since it would be pretty useless to reparse them everytime at startup it would be much nicer to parse them and then serialize the objects to disk so we only need to unserialize the structures. This is in alignment with #47 

The stubs are licensed under Apache2. Does anyone know if that is conflicting with ISC?

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -29,8 +29,27 @@
         "sabre/event": "^4.0",
         "felixfbecker/advanced-json-rpc": "^1.2",
         "squizlabs/php_codesniffer" : "^2.7",
-        "symfony/debug": "^3.1"
+        "symfony/debug": "^3.1",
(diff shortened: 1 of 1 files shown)


felixfbecker:
https://github.com/JetBrains/phpstorm-stubs/pull/125

felixfbecker:
https://packagist.org/packages/phpstubs/phpstubs

gza:
https://packagist.org/packages/phpstubs/phpstubs is pretty outdated, no ?

felixfbecker:
Yes it is, I was just collecting alternatives

Issue #57, opened 2016-10-10
Indexing became extremely slow


mniewrzal:
After recent changes indexing performance decreased and memory consumption increased extremely. After indexing 3k-4k files slow down is more visible. 

Result for magento (24k files) project:
`All PHP files parsed in 2904 seconds. 6123 MiB allocated.`
Previous those values were about 100s and 600MiB.

felixfbecker:
That really is a lot. Suggestions?

kaloyan-raev:
It would be nice to have some configurable memory limit. And have the language server smart enough to fit in this memory limit, i.e. If the memory limit is tiny, the language server optimize the memory usage by moving this to a file system cache.

Our use case with Eclipse Che, is that the PHP language server runs as an agent in a Docker container with a total memory of 1-2 GB. There are more language severs (JSON, CSS, JavaScript, etc.), Che workspace agent, etc. I would imagine that the PHP language server in our case should be limited to 100-200 MB of memory.

mniewrzal:
Regarding memory limit I think it would be nice to add command line argument like "--tcp" for socket connection and apply it to `ini_set` in bin/php-language-server.php.

As for the performance problem I will take a look what is causing such slow down.

kaloyan-raev:
We have some experience with indexing in Eclipse PDT / Zend Studio. It was impossible to hold all data in-memory. Even a single regular project using a popular framework go over 10,000 files. The nature of the Eclipse workspace allows multiple project at the same time and the memory usage goes really crazy.

For long years we used an H2 database to store the data. H2 is an embedded relational database and as such it has some performance limitations. In the last year we switched to using Apache Lucene - a text-based search engine. It improve the performance significantly. Currently, Eclipse PDT has a much better performance than we have here, even without the latest changes which caused the big slow down.

felixfbecker:
Moving structures to disk would help.

Some ideas to remove RAM usage:
- I implemented the definition map at project level so that a reference to the `PhpDocument` is saved, not the node. An improvement would be to discard the AST, and reparse the document upon request for definition/reference.
- `$content` is currently kept in memory _in addition_ to the AST. Which means we basically keep all file contents in memory. It is only used for formatting, so I propose we re-read the content on formatting request.

Any ideas regarding index time? We traverse all ASTs twice at index time.

felixfbecker:
@kaloyan-raev I heard from other folks that rebuilding the Eclipse PDT index takes them ~3min.
[...]

PR #60, opened 2016-10-10
Only hold content for open files in memory

felixfbecker:
Addressing https://github.com/felixfbecker/php-language-server/issues/57#issuecomment-252572439

Content is only set on `didOpen`, and removed on `didClose`.
For indexing only the AST is kept.

It didn't reduce RAM usage unfortunately.

### DIFF
--- a/src/LanguageServer.php
+++ b/src/LanguageServer.php
@@ -153,7 +153,7 @@ class LanguageServer extends \AdvancedJsonRpc\Dispatcher
                 $shortName = substr($file, strlen($rootPath) + 1);
                 $this->client->window->logMessage(MessageType::INFO, "Parsing file $fileNum/$numTotalFiles: $shortName.");
 
-                $this->project->getDocument($uri)->updateContent(file_get_contents($file));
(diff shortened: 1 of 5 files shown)


mniewrzal:
I think you need unset $statements too with removeContent(). Memory saving for magento after removing content from PhpDocument on close what just about 20% (630MiB => 500 MiB).

felixfbecker:
Yeah but that will not work with current implementation of definition index, they reference AST node, which in turn references parent node, etc.

20% is at least something.

I will look into completely removing the AST and only keeping an index of FQN => PhpDocument, then reparse the document on demand.

felixfbecker:
Can we merge #35? Need to use `uriToPath` function

mniewrzal:
> Can we merge #35? Need to use uriToPath function

Yes, will you do rebase?

felixfbecker:
Actually let's merge this first, doesn't hurt. Any objections?

mniewrzal:
Nothing from my side.

PR #61, opened 2016-10-10
Don't keep AST in memory, lazy load content on demand

felixfbecker:
Extension of #60 

Now only a FQN => PhpDocument map is kept at the project level.
Upon a definition request the content/AST of the document is lazily loaded from disk.

Indexing is still slow but RAM usage is now as low as <100MB even for magento (for some reason though VS Code crashed after ~20,000 files).

I moved all LSP-related logic like `findSymbols` out of `Project` and `PhpDocument` into the method handlers. `Project` and `PhpDocument` in turn provide methods like `getDefinition` or `getDefinitionDocument`.

workspace/symbol now also searches over the FQN, which is a nice side-effect.
The downside though is that workspace/symbol now needs to reparse every document matching the query. So in the case that your query is not very specific, it will take forever in large projects and bring your RAM usage up to the sky.

Test with Symfony (7100 files):
```
All PHP files parsed in 259 seconds. 52 MiB allocated.
```


mniewrzal:
With this PR I have reasonable indexing times:
magento
```
[Info  - 9:57:14 PM] Parsing file 24273/24273: dev/tests/functional/tests/app/Magento/Email/Test/Constraint/AssertEmailTemplateSuccessSaveMessage.php.
[Info  - 9:57:14 PM] All PHP files parsed in 119 seconds. 575 MiB allocated.
``` 
symfony
```
[Info  - 9:58:36 PM] Parsing file 7099/7100: vendor/zendframework/zend-code/test/Generic/Prototype/PrototypeClassFactoryTest.php.
[Info  - 9:58:36 PM] Parsing file 7100/7100: .github/build-packages.php.
[Info  - 9:58:36 PM] All PHP files parsed in 27 seconds. 50 MiB allocated.
```
To be clear I'm testing with XDebug disabled.

PR #62, opened 2016-10-10
Fixes #59: Handle correctly negative endLine in PHP Parser errors

### LINKED ISSUE #59: Notification handler 'textDocument/publishDiagnostics' failed with message: Illegal argument: line must be positive
Seems like we publish errors as diagnostics that don't have a line, which is kind of pointless

kaloyan-raev:
Fixes #59

### DIFF
--- a/src/PhpDocument.php
+++ b/src/PhpDocument.php
@@ -178,7 +178,7 @@ class PhpDocument
             $diagnostic = new Diagnostic();
             $startLine = max($error->getStartLine() - 1, 0);
             $startColumn = $error->hasColumnInfo() ? $error->getStartColumn($this->content) - 1 : 0;
-            $endLine = $error->getEndLine() === -1 ? $startLine : $error->getEndLine() - 1;
+            $endLine = max($error->getEndLine() - 1, $startLine);
             $endColumn = $error->hasColumnInfo() ? $error->getEndColumn($this->content) : 0;
             $diagnostic->range = new Range(new Position($startLine, $startColumn), new Position($endLine, $endColumn));
             $diagnostic->severity = DiagnosticSeverity::ERROR;
--- a/src/PhpDocument.php
+++ b/src/PhpDocument.php
@@ -176,10 +176,11 @@ class PhpDocument
         $diagnostics = [];
         foreach ($errors as $error) {
             $diagnostic = new Diagnostic();
-            $diagnostic->range = new Range(
-                new Position($error->getStartLine() - 1, $error->hasColumnInfo() ? $error->getStartColumn($this->content) - 1 : 0),
-                new Position($error->getEndLine() - 1, $error->hasColumnInfo() ? $error->getEndColumn($this->content) : 0)
-            );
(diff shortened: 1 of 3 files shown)


src/PhpDocument.php
```diff
@@ -176,10 +176,11 @@ public function parse()
         $diagnostics = [];
         foreach ($errors as $error) {
             $diagnostic = new Diagnostic();
-            $diagnostic->range = new Range(
-                new Position($error->getStartLine() - 1, $error->hasColumnInfo() ? $error->getStartColumn($this->content) - 1 : 0),
-                new Position($error->getEndLine() - 1, $error->hasColumnInfo() ? $error->getEndColumn($this->content) : 0)
-            );
+            $startLine = $error->getStartLine() === -1 ? 0 : $error->getStartLine() - 1;
```
felixfbecker: Why not
```php
max($error->getStartLine() - 1, 0)
```

kaloyan-raev: Agree.

src/PhpDocument.php
```diff
@@ -176,7 +176,7 @@ public function parse()
         $diagnostics = [];
         foreach ($errors as $error) {
             $diagnostic = new Diagnostic();
-            $startLine = $error->getStartLine() === -1 ? 0 : $error->getStartLine() - 1;
+            $startLine = max($error->getStartLine() - 1, 0);
             $startColumn = $error->hasColumnInfo() ? $error->getStartColumn($this->content) - 1 : 0;
             $endLine = $error->getEndLine() === -1 ? $startLine : $error->getEndLine() - 1;
```
felixfbecker: Wouldn't the same work here?
```php
max($error->getEndLine() - 1, $startLine)
```

kaloyan-raev: I wondered about the end line too... In this patch I just wanted to protect from `-1` line values. Your suggestion is to protect `endLine` being lower than `startLine`. Which is logical, but I am not sure if we would not cause some side effect in edge cases.

felixfbecker: Well one possible point of error you leave open is if it returns 0 (or `null`/`false`), because then we subtract 1 from it, and have `-1` again. I think the `max` approach is the safest.

kaloyan-raev: You are right. I made the change in the last commit.

tests/Server/TextDocument/ParseErrorsTest.php
```diff
@@ -38,18 +32,22 @@ public function publishDiagnostics(string $uri, array $diagnostics)
...
         };
-
         $project = new Project($client);
-
-        $textDocument = new Server\TextDocument($project, $client);
-
-        // Trigger parsing of source
+        $this->textDocument = new Server\TextDocument($project, $client);
+    }
+
```
felixfbecker: Trailing whitespace
I recommend you to use an EditorConfig extension 😉

kaloyan-raev: Done.

PR #63, opened 2016-10-10
Lazy load alternative

felixfbecker:
Yet another extension of #61 

Initially I thought keeping the AST in memory would be okay, and we only remove the content. This didn't save much, so I also removed all the nodes. What I ended up with was a lot of code around `PhpDocument`s to manage if the content/AST is loaded: `isLoaded()`, `load()`, `unload()`, if statements and exceptions everywhere.

This simplifies it by making a `PhpDocument` always have all content/nodes (like before).
Only now the `PhpDocument`s are only stored in the `Project` for opened files. For the rest of the project, we create just load it once, add the definitions to the index and save the document URI, then let the object get garbage collected again.


mniewrzal:
From performance perspective this PR is also working ok for me. Magento indexing:
```
[Info  - 11:10:06 AM] Parsing file 24273/24273: dev/tests/functional/tests/app/Magento/Email/Test/Constraint/AssertEmailTemplateSuccessSaveMessage.php.
[Info  - 11:10:06 AM] All PHP files parsed in 121 seconds. 571 MiB allocated.
```

Issue #55, opened 2016-10-10
Exceptions vs returning null


felixfbecker:
We currently return `null` or empty arrays in many places if an action could not be performed. Examples:
- Definition for a position that could not be mapped to a Node
- Definition for a node we couldn't find the definition for
- Any request for a document that is not parsed yet / invalid (`$stmts` is not available)

This results in a lot of repeated `if` checks, not covered branches, possible NullPointers, and potential violation of the protocol (when `null` is not allowed as a return value for `result`)

An Exception would bubble and get turned into an ErrorResponse, which probably results in an error message in the editor. I would like to find out how other language servers do this. We could still catch exceptions and return null in the method handler.

Opinions?

PR #64, opened 2016-10-11
Return empty array instead of null for empty definitions result

kaloyan-raev:
The result of the [Definition Request](https://github.com/Microsoft/language-server-protocol/blob/master/protocol.md#goto-definition-request) is `Location|Location[]`. LSP says nothing about `null`. Thus, if no definition is found we should return an empty array instead of null.

Returning null breaks at least one client - the Typefox Java binding.

### DIFF
--- a/src/Server/TextDocument.php
+++ b/src/Server/TextDocument.php
@@ -96,18 +96,18 @@ class TextDocument
      *
      * @param TextDocumentIdentifier $textDocument The text document
      * @param Position $position The position inside the text document
-     * @return Location|Location[]|null
+     * @return Location|Location[]
      */
     public function definition(TextDocumentIdentifier $textDocument, Position $position)
     {
         $document = $this->project->getDocument($textDocument->uri);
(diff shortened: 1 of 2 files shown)


felixfbecker:
True, violating the protocol here. I was wondering whether it would be better to through an exception though, see #55 . Do you know what other language servers do?

felixfbecker:
Seems like the TS server does not return an error, so I'm fine with it

kaloyan-raev:
>  I was wondering whether it would be better to through an exception though, see #55

No definition found should be a valid situation and should not be considered as an error state.

kaloyan-raev:
> Seems like the TS server does not return an error, so I'm fine with it

@felixfbecker Could you point me to the source code of the TS server? I'd like to examine it for other things.

felixfbecker:
Again it is not open source, maybe not even using LSP yet. I just checked if VS Code displays an error when I get a definition for whitespace.

felixfbecker:
*Afaik

PR #67, opened 2016-10-11
Remove unused use statements

mniewrzal:


felixfbecker:
@kaloyan-raev Please let me review the PR next time too. I know it's a minor change and I certainly don't have any objections, but I prefer squashed commits in most cases instead of a merge commit.

kaloyan-raev:
@felixfbecker OK. Sorry. It was really trivial, so I went ahead and merged it.

felixfbecker:
No problem, it's just the merge commits in master that get annoying. Gonna clean that up really quick with a force push.

kaloyan-raev:
Perhaps, you can disallow "merge commits" from settings and leave only "squash merging".

felixfbecker:
Yeah did that now. Sometimes though I have a PR that has a few commits which should really exist in the history on its own, so I like to have the choice.

PR #68, opened 2016-10-11
Add command line parameter for memory limit #66

### LINKED ISSUE #66: Add command line parameter for memory limit
One of the conclusions from #57 is that it would be nice to add command line param to configure PHP memory limit. I think it shouldn't be a problem to add parameter like `--memory-limit` and use it with `ini_set` in `php-language-server.php`. Similar to existing `--tcp` parameter.

mniewrzal:
If this implementation will be approved then I will prepare PR for vscode-php-intellisense. With `getopt` optional parameter supports only notation with '=' ('--some-param=value' or '--some-param="value"') and because of that `--tcp` param needs to be adjusted. 

Closes #66

### DIFF
--- a/bin/php-language-server.php
+++ b/bin/php-language-server.php
@@ -4,7 +4,9 @@ use LanguageServer\{LanguageServer, ProtocolStreamReader, ProtocolStreamWriter};
 use Sabre\Event\Loop;
 use Symfony\Component\Debug\ErrorHandler;
 
-ini_set('memory_limit', '-1');
+$options = getopt('', ['tcp::', 'memory-limit::']);
+
+ini_set('memory_limit', $options['memory-limit'] ?? -1);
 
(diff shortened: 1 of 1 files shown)


felixfbecker:
Did a quick rebase for you.

felixfbecker:
Reading through the docs it seems like `getopt` supports options delimited by spaces as well:
http://php.net/manual/en/function.getopt.php

bin/php-language-server.php
```diff
@@ -17,8 +19,8 @@
 
 cli_set_process_title('PHP Language Server');
 
-if (count($argv) >= 3 && $argv[1] === '--tcp') {
-    $address = $argv[2];
+if ($options['tcp']) {
```
felixfbecker: I think this should use `isset()` to prevent `Undefined Index` warning

mniewrzal:
Space can be used as a separator only for required parameters. From manual:
`Note: Optional values do not accept " " (space) as a separator.`

PR #70, opened 2016-10-11
Update README with --memory-limit param

mniewrzal:


README.md
```diff
@@ -30,12 +30,17 @@ Run the tests with
...
 
-    --tcp host:port
-
+###### --tcp=host:port (optional)
 Causes the server to use a tcp connection for communicating with the language client instead of using STDIN/STDOUT.
 The server will try to connect to the specified address.
 
 Example:
 
     php bin/php-language-server.php --tcp 127.0.0.1:12345
```
felixfbecker: The `=` is missing here

felixfbecker:
Btw what happens when you run vscode-php-intellisense without your new PR against latest master? Does it break or just not work with TCP? Otherwise we have to do a major version

mniewrzal:
TCP mode will not work under Windows but it should start STDIN/STDOUT communication. I think so :)

PR #76, opened 2016-10-13
Go to definition doesn't work for method return type

### LINKED ISSUE #75: Go to definition doesn't work for method return type
Example:
```php
<?php
namespace TestNamespace;

class MyTest
{
    public function abc() : TestClass // it does not work
    {
        new TestClass(); // it works
    }
}
```

mniewrzal:
Closes #75


felixfbecker:
Adding a new fixture for every bug will become unmanageable. What are your suggestions on improving the tests?

MadHed:
I'm not an expert at unit testing but my gut feeling tells me that a fixture specific to a bug should only be added if the bug appears as the direct result of a corner case in the parsed source code.
For testing normal operation a single shared fixture containing all testable cases should suffice, no?

felixfbecker:
@mniewrzal I'm working on a refactor of the tests which will make it much easier in the future to change fixtures

mniewrzal:
@felixfbecker Ok, nice :)

felixfbecker:
@mniewrzal #87

felixfbecker:
@mniewrzal would you mind rebasing this on the new test structure?

mniewrzal:
I will do this today.

felixfbecker:
Why not use the return type of https://github.com/felixfbecker/php-language-server/blob/master/fixtures/global_references.php#L22 ?

mniewrzal:
Case with function return type was covered earlier - https://github.com/felixfbecker/php-language-server/blob/master/tests/Server/TextDocument/Definition/GlobalTest.php#L141. This is fix for method return type.

felixfbecker:
Okay, sorry. Could you then please just add a return type to one of the methods defined in `symbols.php` so we don't clutter the fixtures with new classes for every new test?

mniewrzal:
Ok

mniewrzal:
Anything I should change or its ready to go? I don't want to resolve conflicts one more time ;)

felixfbecker:
No, looks great :)

mniewrzal:
Cool, thanks for review :)

PR #80, opened 2016-10-13
Add size limit of 0.5MB for indexing

### LINKED ISSUE #71: Add size limit for processed PHP file
Very large files can consume disproportionate huge amount of memory. Composer project contains file LargeClass.php and this one file is bumping memory consumption from 18MiB to 542MiB. Such large files are not real life production code (mostly fixtures) and does not make sense to process them. I believe that limit around 0.5MB will be good enough.

Example of such file:
https://github.com/composer/composer/blob/master/tests/Composer/Test/Autoload/Fixtures/classmap/LargeClass.php

felixfbecker:
Closes #71 
@mniewrzal


mniewrzal:
Works and looks ok for me. Thanks!

PR #82, opened 2016-10-13
Cache index on disk

### LINKED ISSUE #47: Cache indexing result

felixfbecker:
Closes #47

This is a very simple implementation of a disk cache. 
When index finished and on shutdown the definitions and references are saved to `.phpls/definitions.json` and `.phpls/references.json`.
On initialization, they are restored before beginning to reindex.
No fancy cache invalidation.
The result is when you once indexed a project, the second time you open it you can use all features instantly without having to wait for indexing to finish.

Magento:
```
[Info  - 02:07:26] Saving 203562 definitions to cache
[Info  - 02:07:26] Saving 323550 references to cache
```

### DIFF
--- a/src/LanguageServer.php
+++ b/src/LanguageServer.php
@@ -181,6 +181,10 @@ class LanguageServer extends AdvancedJsonRpc\Dispatcher
                     }
                 }
 
+                if ($fileNum % 1000 === 0) {
+                    $this->saveCache();
+                }
+
                 Loop\setTimeout($processFile, 0);
             } else {
                 $duration = (int)(microtime(true) - $startTime);
--- a/src/LanguageServer.php
+++ b/src/LanguageServer.php
@@ -101,11 +101,10 @@ class LanguageServer extends AdvancedJsonRpc\Dispatcher
     {
         $this->rootPath = $rootPath;
 
-        $this->restoreCache();
-
         // start building project index
-        if ($rootPath) {
-            $this->indexProject($rootPath);
+        if ($rootPath !== null) {
+            $this->restoreCache();
+            $this->indexProject();
         }
 
         $serverCapabilities = new ServerCapabilities();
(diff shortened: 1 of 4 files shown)


mniewrzal:
Indexing is still triggered after restoring cache.

src/LanguageServer.php
```diff
@@ -157,14 +167,69 @@ private function indexProject(string $rootPath)
...
 
         Loop\setTimeout($processFile, 0);
     }
+
+    /**
+     * Restores the definition and reference index from the .phpls cache directory, if available
+     *
+     * @return void
+     */
+    public function restoreCache()
```
mniewrzal: I think this method can return true if both 'definitions' and 'references' are restored correctly. If one of them is missing then cache is invalid.

felixfbecker: And what would that return value be used for?

felixfbecker:
@mniewrzal that is intended. It is just for having some feature support in the first minute of indexing there already, but then quickly update the index and cache. A lot can happen to files while the LS is not running, user can switch branches etc. and I don't want to deal with cache invalidation at this point.

mniewrzal:
Ok, I understand your conception now. Ignore my comment to restoreCache method :)

Works ok on my machine :)

felixfbecker:
Thanks for testing. Anyone else objections? @kaloyan-raev @levan-g @MadHed

kaloyan-raev:
No objections. If it's ok for @mniewrzal then it is ok for me too :)

MadHed:
Works for me. Is there a standard for where to store per-workspace extension metadata?
I thought these would go into .vscode.

mniewrzal:
> On second thought... since this project will be used by other IDEs maybe .vscode would not be so great :D

Exactly ;)

PR #84, opened 2016-10-13
Add support for constants with define()

### LINKED ISSUE #48: Support define() constants
Parsed as a function call node

felixfbecker:
Closes #48

### DIFF
--- a/src/Protocol/SymbolInformation.php
+++ b/src/Protocol/SymbolInformation.php
@@ -58,13 +58,25 @@ class SymbolInformation
             Node\Stmt\PropertyProperty::class => SymbolKind::PROPERTY,
             Node\Const_::class                => SymbolKind::CONSTANT
         ];
-        $class = get_class($node);
-        if (!isset($nodeSymbolKindMap[$class])) {
-            throw new Exception("Not a declaration node: $class");
-        }
         $symbol = new self;
-        $symbol->kind = $nodeSymbolKindMap[$class];
-        $symbol->name = (string)$node->name;
+        if (
+            $node instanceof Node\Expr\FuncCall
+            && $node->name instanceof Node\Name
+            && (string)$node->name === 'define'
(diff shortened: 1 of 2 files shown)


jens1o:
I think we could run into another problem. https://github.com/felixfbecker/php-language-server/blob/master/src/LanguageServer.php#L283 We would need to watch for every little char, won't we?

felixfbecker:
No, those are just the "explicit" trigger characters.

felixfbecker:
Sure, go ahead! be aware though this has quite some merge conflicts now

jens1o:
So I'm creating a new pr with the tests?

felixfbecker:
Yep

src/Protocol/SymbolInformation.php
```diff
@@ -58,13 +58,25 @@ public static function fromNode(Node $node, string $fqn = null)
...
-        if (!isset($nodeSymbolKindMap[$class])) {
-            throw new Exception("Not a declaration node: $class");
-        }
         $symbol = new self;
-        $symbol->kind = $nodeSymbolKindMap[$class];
-        $symbol->name = (string)$node->name;
+        if (
+            $node instanceof Node\Expr\FuncCall
+            && $node->name instanceof Node\Name
+            && (string)$node->name === 'define'
```
staabm: Will this false positive on a namespaced function named `define`?

jens1o: I'll check this, but it should be excluded with `$node->name instanceof Node\Name`

jens1o: It doesn't care about this, the way I implemented it is right.

felixfbecker:
@jens1o are you still working on this?

jens1o:
It's finished, I just forgot to create the pr. I'll do it now. Thanks for the ping!

PR #89, opened 2016-10-18
Instantiate Project in intialize()

felixfbecker:
This should allow to re-initialize the LS for a different project without restarting it

@mniewrzal is this what you had in mind?


mniewrzal:
Yes, should be ok right now :+1:

PR #104, opened 2016-10-20
Switch to serialize() instead of JSON

felixfbecker:
There is no reason really for using JSON, especially since we need to serialize classes now. It is much easier to use native `serialize()`/`unserialize()`.

### DIFF
--- a/src/LanguageServer.php
+++ b/src/LanguageServer.php
@@ -207,16 +207,14 @@ class LanguageServer extends AdvancedJsonRpc\Dispatcher
     {
         $cacheDir = $this->rootPath . '/.phpls';
         if (is_dir($cacheDir)) {
-            if (file_exists($cacheDir . '/symbols.json')) {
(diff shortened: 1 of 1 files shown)


MadHed:
One thing to keep in mind though when using serialize/unserialize: The serialized format is bound to the class structure at creation time. Should the classes ever change (add/remove/rename fields) the serialized data will not recreate the original objects.

felixfbecker:
Good point, but when we change the class structure, probably the cache format will have to change too

MadHed:
True, also since currently the only serialized class is SymbolInformation which is probably not going to change this should be ok.

We should keep this in mind when the serialization becomes more advanced and add a version field that lets us decide if we need to recreate the cache or not.

felixfbecker:
Btw I don't think it will break. If the class name stays the same, it should work. Otherwise it will be unserialized as `__PHP_Incomplete_Class`. But in the end I would prefer to save a version in a meta file and nuke the cache if the version changed.

PR #107, opened 2016-10-22
Linting

ADmad:
Refs #12

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,7 +31,7 @@ script:
 
  - if [[ $DEFAULT = 1 && $TRAVIS_PHP_VERSION != 7.0 ]]; then vendor/bin/phpunit; fi
 
- - if [[ $PHPCS = 1 ]]; then vendor/bin/phpcs -p --extensions=php --standard=PSR2 ./src ./tests; fi
+ - if [[ $PHPCS = 1 ]]; then vendor/bin/phpcs -p --extensions=php --standard=phpcs.xml.dist ./src ./tests; fi
 
 after_success:
  - if [[ $DEFAULT = 1 && $TRAVIS_PHP_VERSION = 7.0 ]]; then bash <(curl -s https://codecov.io/bash); fi
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,40 +1,18 @@
 language: php
 
-sudo: false
-
-dist: trusty
-
 php:
  - '7.0'
- - '7.1'
-
-env:
-  global:
-    - DEFAULT=1
-
-matrix:
-  fast_finish: true
-
-  include:
-    - php: '7.0'
-      env: PHPCS=1 DEFAULT=0
-
-  allow_failures:
-    - env: PHPCS=1 DEFAULT=0
 
 cache:
   directories:
    - vendor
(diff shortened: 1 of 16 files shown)


ADmad:
There's one pending coding standard [error](https://travis-ci.org/ADmad/php-language-server/jobs/169792734#L186) which I wasn't sure how to take care of, so please look into that.

ADmad:
The coding standards used is [PSR2](http://www.php-fig.org/psr/psr-2/) which is the widely accepted coding style guide for PHP.

ADmad:
> Group use statements

Yes it's a PSR2 requirement "There MUST be one use keyword per declaration."
Having one `use` declaration per line has it's benefits. For e.g. the Atom editor has a plugin which auto adds `use` statement at top of file when you use `new Foo`. I wish one day a vscode plugin would do the same :)

> The line limit of 80 characters is just straight from the 90s. We all have 16:9 monitors

It's not 80, PSR2 has a soft limit of 120 characters.

> Is there a way to ignore the space errors in the definition locations in the tests so they can stay table-like?

Yes you have use `// @ignoreStandardsStart` and `// @ignoreStandardsEnd` tag to make phpcs ignore violations in code lines between those tags. It have used it at couple of places in tests as I didn't want to go through the trouble of manually breaking long lines to under the 120 lines limit :)

Most of formatting changes were done automatically using `phpcbf` which comes along with `phpcs`.

> Again you can configure phpcs with an XML file, so that would be appreciated.

Yes, I'll add it.

ADmad:
I have added `phpcs.xml.dist`. You can see a good example of how you can customize the rules [here](https://github.com/cakephp/cakephp-codesniffer/blob/master/CakePHP/ruleset.xml).

ADmad:
Having one `use` per declaration also allows sniffers to easily check for unused declarations.

felixfbecker:
I don't see why group use statements wouldn't let the tools do this, at least for this LS I plan to fully support them. It's a new feature so it may not be widely supported but it makes the code so much more readable, and that's what linting is all about, checking that code style is consistent so it's easier to read 😄

mniewrzal:
It looks that for now this behavior wont be changed ([PHP_CodeSniffer #878](https://github.com/squizlabs/PHP_CodeSniffer/issues/878)) because PSR2 standard is fixed and not adjusted to PHP 7. I think reasonable solution would be to exclude this rule with configuration file.

ADmad:
You guys can follow the coding style that suits you :) 

You can just take the updated `.travis.yml` and new `phpcs.xml.dist` from this PR, update the latter as per your needs and then use `phpcbf` to autofix the coding style violations.

felixfbecker:
@ADmad will do 😉 Thanks for your help

ADmad:
No problem, really appreciate the work you are doing to make PHP language support awesome in vscode.

ADmad:
I have undone all code changes.

mniewrzal:
When this PR will be merged I can prepare next with adjusted config file to exclude things like use statements rule (and others if we will need).

felixfbecker:
Thanks a lot @ADmad!

PR #110, opened 2016-10-24
Exit process if input or output stream is broken

### LINKED ISSUE #106: The language server should exit if the parent process dies
According to LSP:

> If the parent process is not alive then the server should exit (see exit notification) its process.

Currently if the parent process exits without sending the exit notification, the process of the language server goes in a high CPU loop. While experimenting, I often fall in this situation and have to kill the zombie processes manually.

kaloyan-raev:
Fixes #106.

This is a simple patch that makes the language server to exit if the parent process died.

1. The custom error handling in the `ProtocolStreamWriter` is removed, so `fwrite`-ing to a closed stream results in throwing an exception. The process exits because of it.
2. An EOF check is added to the `ProtocolStreamReader`. If `stream_select()` activates the `ProtocolStreamReader` and there is only EOF in the stream then, most probably, the stream is already closed.

This PR can be improved, but for now I am pushing it like this to discuss if this is a good approach. It works for me at least. I tested different scenarios and in all of them if the VSCode process exits, the language server process exits too.

Possible ways to improve the new check in the `ProtocolStreamReader` would be:

1. Call `shutdown()` and `exit()` instead of just `die`.
2. if `feof()` returns `true` then do an additional check if the parent PID has really gone.

### DIFF
--- a/src/ProtocolStreamReader.php
+++ b/src/ProtocolStreamReader.php
@@ -31,7 +31,7 @@ class ProtocolStreamReader implements ProtocolReader
         $this->input = $input;
         Loop\addReadStream($this->input, function() {
             if (feof($this->input)) {
-                throw new RuntimeException("Stream is closed.");
+                throw new RuntimeException('Stream is closed');
             }
 
             while (($c = fgetc($this->input)) !== false && $c !== '') {
--- a/src/ProtocolStreamReader.php
+++ b/src/ProtocolStreamReader.php
@@ -6,6 +6,7 @@ namespace LanguageServer;
 use LanguageServer\Protocol\Message;
 use AdvancedJsonRpc\Message as MessageBody;
 use Sabre\Event\Loop;
+use RuntimeException;
 
 abstract class ParsingMode
 {
@@ -30,7 +31,7 @@ class ProtocolStreamReader implements ProtocolReader
         $this->input = $input;
         Loop\addReadStream($this->input, function() {
             if (feof($this->input)) {
-                die;
+                throw new RuntimeException("Stream is closed.");
             }
 
             while (($c = fgetc($this->input)) !== false && $c !== '') {
--- a/src/ProtocolStreamReader.php
+++ b/src/ProtocolStreamReader.php
@@ -29,6 +29,10 @@ class ProtocolStreamReader implements ProtocolReader
     {
         $this->input = $input;
         Loop\addReadStream($this->input, function() {
+            if (feof($this->input)) {
+                die;
+            }
+
             while (($c = fgetc($this->input)) !== false && $c !== '') {
                 $this->buffer .= $c;
                 switch ($this->parsingMode) {
--- a/src/ProtocolStreamWriter.php
+++ b/src/ProtocolStreamWriter.php
@@ -31,16 +31,7 @@ class ProtocolStreamWriter implements ProtocolWriter
         $totalBytesWritten = 0;
 
         while ($totalBytesWritten < $msgSize) {
-            error_clear_last();
-            $bytesWritten = @fwrite($this->output, substr($data, $totalBytesWritten));
-            if ($bytesWritten === false) {
-                $error = error_get_last();
-                if ($error !== null) {
-                    throw new RuntimeException('Could not write message: ' . error_get_last()['message']);
-                } else {
-                    throw new RuntimeException('Could not write message');
-                }
-            }
+            $bytesWritten = fwrite($this->output, substr($data, $totalBytesWritten));
             $totalBytesWritten += $bytesWritten;
         }
     }
--- a/src/ProtocolStreamWriter.php
+++ b/src/ProtocolStreamWriter.php
@@ -4,7 +4,6 @@ declare(strict_types = 1);
 namespace LanguageServer;
 
 use LanguageServer\Protocol\Message;
-use RuntimeException;
 
 class ProtocolStreamWriter implements ProtocolWriter
 {


src/ProtocolStreamReader.php
```diff
@@ -29,6 +29,10 @@ public function __construct($input)
     {
         $this->input = $input;
         Loop\addReadStream($this->input, function() {
+            if (feof($this->input)) {
+                die;
```
felixfbecker: Never ever simply die in a function if an error condition happens. Throw an exception, and if it is not caught, it will result in the process exiting with a non-zero exit code.

felixfbecker: Never ever simply die in a function if an error condition happens. Throw an exception, and if it is not caught, it will result in the process exiting with a non-zero exit code.

src/ProtocolStreamWriter.php
```diff
@@ -31,16 +31,7 @@ public function write(Message $msg)
...
-            $bytesWritten = @fwrite($this->output, substr($data, $totalBytesWritten));
-            if ($bytesWritten === false) {
-                $error = error_get_last();
-                if ($error !== null) {
-                    throw new RuntimeException('Could not write message: ' . error_get_last()['message']);
-                } else {
-                    throw new RuntimeException('Could not write message');
-                }
-            }
+            $bytesWritten = fwrite($this->output, substr($data, $totalBytesWritten));
```
felixfbecker: 100% agree with this change. I'm not even sure if the error handling worked before. It was written when we didn't have an ErrorHandler to convert notices to exceptions.

kaloyan-raev:
> Regarding the feof check though, could you explain a bit more why it makes sense? Would addReadStream ever call the callback if the stream is EOF?

If the input stream is broken then `stream_select()` activates the `ProtocolStreamReader` with only EOF in the input stream. I found this using debugging.

> Did you test if your bug occurs only with the removal of the error handling or is the feof check a crucial part of your fix?

Initially I tried removing only the error handling in the `ProtocolStreamWriter`. It was sufficient as long as the language server has anything to write to the client, i.e. a log message during indexing. But if I kill VSCode during a calm situation then the event loop starts cycling on the `ProtocolStreamReader`. See above.

src/ProtocolStreamReader.php
kaloyan-raev: OK. I throw RuntimeException now.

felixfbecker:
@kaloyan-raev With this PR I now get an error when doing a symbol search while indexing is running for magento:

```
PHP Fatal error:  Uncaught Symfony\Component\Debug\Exception\ContextErrorException: Notice: fwrite(): send of 8192 bytes failed with errno=10035 A non-blocking socket operation could not be executed immediately.
 in C:\Users\felix\git\OpenSource\php-language-server\src\ProtocolStreamWriter.php:33
Stack trace:
#0 [internal function]: Symfony\Component\Debug\ErrorHandler->handleError(8, 'fwrite(): send ...', 'C:\\Users\\felix\\...', 33, Array)
#1 C:\Users\felix\git\OpenSource\php-language-server\src\ProtocolStreamWriter.php(33): fwrite(Resource id #16, 'Content-Type: a...')
#2 C:\Users\felix\git\OpenSource\php-language-server\src\LanguageServer.php(80): LanguageServer\ProtocolStreamWriter->write(Object(LanguageServer\Protocol\Message))
#3 C:\Users\felix\git\OpenSource\php-language-server\src\ProtocolStreamReader.php(53): LanguageServer\LanguageServer->LanguageServer\{closure}(Object(LanguageServer\Protocol\Message))
#4 C:\Users\felix\git\OpenSource\php-language-server\vendor\sabre\event\lib\Loop\Loop.php(311): Languag in C:\Users\felix\git\OpenSource\php-language-server\src\ProtocolStreamWriter.php on line 33

**Comment by kaloyan-raev** [comment]
I found this that might be related: https://bugs.php.net/bug.php?id=3906

**Comment by felixfbecker** [comment]
How is that bug related?

**Comment by kaloyan-raev** [comment]
Sorry! Copy-paste mistake. I edited my previous comment with the correct URL.

**Comment by felixfbecker** [comment]
We had this before here: https://github.com/felixfbecker/php-language-server/pull/31#issuecomment-250546775

**Comment by felixfbecker** [comment]
The problem here is that this is a `Notice` which is actually not a problem because we buffer the writing, but because this PR removed the `@` operator it is now thrown as an `ErrorException`. I am thinking about whether we can somehow catch this notice but still let your error throw?

**Comment by kaloyan-raev** [comment]
In the case of broken pipe the error is again an `ErrorException` with message:

```
Notice: fwrite(): send of 298 bytes failed with errno=32 Broken pipe
```

I guess the message is the only difference between the two cases.

**Comment by kaloyan-raev** [comment]
So, I am not sure if `errno=10035` is not really a problem. As you can see even `errno=32 Broken pipe` is a Notice although it is a real problem.

**Comment by MadHed** [comment]
`errno=10035` is normal behaviour when the stream is non blocking and we are writing more data than fits into the outgoing stream buffer at the moment.

The real wtf here is why `errno=32 Broken pipe` is a notice and not an error...

**Comment by kaloyan-raev** [comment]
Can we split the data to be written into chunks less than 8192 bytes and avoid `errno=10035`?

**Comment by felixfbecker** [comment]
Can we get the `errno` from the `ErrorException` if we catch it?

**Comment by MadHed** [comment]
I'm not sure if that would be a permanent fix or just hide the problem... Theoretically `errno=10035` can also appear if we are writing data faster than the IDE can consume them.

**Comment by kaloyan-raev** [comment]
> Can we get the errno from the ErrorException if we catch it?

As far as I can see, only if we parse the exception's message.

**Comment by felixfbecker** [comment]
Yeah but `fwrite` returns the bytes written, we already handle that

**Comment by MadHed** [comment]
> Yeah but fwrite returns the bytes written, we already handle that

I was replying to this message:

> Can we split the data to be written into chunks less than 8192 bytes and avoid errno=10035?

**Comment by felixfbecker** [comment]
`$e->getCode()` does _not_ give the code. But I don't know if parsing the notice message is a good idea, are these the same on every OS, locale and version?

**Comment by felixfbecker** [comment]
Obviously not

> WSAEWOULDBLOCK
> 10035
> 
> Resource temporarily unavailable.
> This error is returned from operations on nonblocking sockets that cannot be completed immediately, for example recv when no data is queued to be read from the socket. It is a nonfatal error, and the operation should be retried later. It is normal for WSAEWOULDBLOCK to be reported as the result from calling connect on a nonblocking SOCK_STREAM socket, since some time must elapse for the connection to be established.

https://msdn.microsoft.com/en-us/library/windows/desktop/ms740668(v=vs.85).aspx

**Comment by felixfbecker** [comment]
FYI https://github.com/reactphp/stream/blob/master/src/Buffer.php#L90

**Comment by kaloyan-raev** [comment]
Using a custom error handler, I get the following info for `errno=32`:

```
errno = 8
errstr = fwrite(): send of 290 bytes failed with errno=32 Broken pipe
errfile = /home/kraev/git/php-language-server/src/ProtocolStreamWriter.php
errline = 35
```

@felixfbecker What do you get for `errno=10035`?

felixfbecker:
As I said, you cannot get the errno. The code 8 just means `Notice`

kaloyan-raev:
How about the following approach. Let's add back the `@` operator in front of `fwrite`. If `@fwrite` returns 0 then we exit the `ProtocolStreamWriter::write()` method to return back to the event loop. If the issue was `errno=32` then the event loop will wake up the `ProtocolStreamReader`, which will exit the process. But, if the issue was `error=10035` or other then the event loop would make a new attempt to write the data to the client. As far as I understand, the latter is currently not possible, because writing to the client is currently not controlled by the event loop. But there is a plan to change this, right?

MadHed:
> How about the following approach [...]

Yeah, that sounds like the best solution to me. Making the writer async should require no more than to make `write()` append the message to an internal buffer, and call `Loop\addWriteStream()` from the constructor and do the actual writing in the callback.

felixfbecker:
> then the event loop will wake up the ProtocolStreamReader, which will exit the process
@kaloyan-raev what you mean by that?

I am starting to think about if it makes sense to use an object oriented stream abstraction instead.

MadHed:
> the event loop will wake up the ProtocolStreamReader, which will exit the process [...]

Hm... now thinking about it, this only happens when there is data available to read, which is probably not the case when the stream got closed.

kaloyan-raev:
>> then the event loop will wake up the ProtocolStreamReader, which will exit the process

>@kaloyan-raev what you mean by that?

@felixfbecker I meant the execution will reach this line: https://github.com/felixfbecker/php-language-server/blob/master/src/ProtocolStreamReader.php#L31

kaloyan-raev:
> Hm... now thinking about it, this only happens when there is data available to read, which is probably not the case when the stream got closed.

@MadHed Check this comment: https://github.com/felixfbecker/php-language-server/pull/110#issuecomment-255711936

PR #112, opened 2016-10-24
Async ProtocolStreamWriter

MadHed:
Here is my proposal for an async ProtocolStreamWriter.

Currently 1 test fails, because it expects to write all data immediately, which is now not the case anymore.

Maybe this will fix https://github.com/felixfbecker/php-language-server/pull/110#issuecomment-255836119

### DIFF
--- a/src/ProtocolStreamWriter.php
+++ b/src/ProtocolStreamWriter.php
@@ -34,13 +34,15 @@ class ProtocolStreamWriter implements ProtocolWriter
      * Sends a Message to the client
      *
      * @param Message $msg
-     * @return void
+     * @return Promise Resolved when the message has been fully written out to the output stream
      */
     public function write(Message $msg)
     {
         // if the message queue is currently empty, register a write handler.
         if (empty($this->messages)) {
-            Loop\addWriteStream($this->output, [$this, 'writeData']);
+            Loop\addWriteStream($this->output, function () {
+                $this->flush();
+            });
         }
 
         $promise = new Promise();
@@ -53,11 +55,10 @@ class ProtocolStreamWriter implements ProtocolWriter
 
     /**
      * Writes pending messages to the output stream.
-     * Must be public to be able to be used as a callback.
      *
      * @return void
      */
-    public function writeData()
+    private function flush()
     {
         $keepWriting = true;
         while ($keepWriting) {
--- a/src/ProtocolStreamWriter.php
+++ b/src/ProtocolStreamWriter.php
@@ -59,36 +59,33 @@ class ProtocolStreamWriter implements ProtocolWriter
      */
     public function writeData()
     {
-        $message = $this->messages[0]['message'];
-        $promise = $this->messages[0]['promise'];
+        $keepWriting = true;
+        while ($keepWriting) {
+            $message = $this->messages[0]['message'];
+            $promise = $this->messages[0]['promise'];
 
-        error_clear_last();
-        $bytesWritten = @fwrite($this->output, $message);
+            error_clear_last();
+            $bytesWritten = @fwrite($this->output, $message);
 
-        if ($bytesWritten === false) {
-            $error = error_get_last();
-            $promise->reject($error);
-            if ($error !== null) {
-                throw new RuntimeException('Could not write message: ' . error_get_last()['message']);
+            if ($bytesWritten > 0) {
+                $message = substr($message, $bytesWritten);
+            }
+
+            // Determine if this message was completely sent
+            if (strlen($message) === 0) {
+                array_shift($this->messages);
+
+                // This was the last message in the queue, remove the write handler.
+                if (count($this->messages) === 0) {
+                    Loop\removeWriteStream($this->output);
+                    $keepWriting = false;
+                }
+
+                $promise->fulfill();
             } else {
-                throw new RuntimeException('Could not write message');
+                $this->messages[0]['message'] = $message;
+                $keepWriting = false;
             }
-        } else if ($bytesWritten > 0) {
-            $message = substr($message, $bytesWritten);
-        }
-
-        // Determine if this message was completely sent
-        if (strlen($message) === 0) {
-            array_shift($this->messages);
-
-            // This was the last message in the queue, remove the write handler.
-            if (count($this->messages) === 0) {
-                Loop\removeWriteStream($this->output);
-            }
-
-            $promise->fulfill();
-        } else {
-            $this->messages[0]['message'] = $message;
         }
     }
 }
--- a/src/ProtocolStreamWriter.php
+++ b/src/ProtocolStreamWriter.php
@@ -64,7 +64,6 @@ class ProtocolStreamWriter implements ProtocolWriter
             $message = $this->messages[0]['message'];
             $promise = $this->messages[0]['promise'];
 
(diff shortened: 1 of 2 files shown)


felixfbecker:
I think `write()` should return a promise that resolves when the data has been written.

What about this case:
 - Loop says "hey the stream is ready, you can write stuff" - buffer is empty though
 - `write()` is called
 - when will the callback be called?

MadHed:
> I think write() should return a promise that resolves when the data has been written.

Good Idea.

> when will the callback be called by the Loop?

The callback will be called every time `stream_select()` is called... which I just found out. When doing nothing the server just sits there busy-waiting until a message arrives.

felixfbecker:
> The callback will be called every time stream_select() is called... which I just found out. When doing nothing the server just sits there busy-waiting until a message arrives, consuming cpu cycles.

"Until a message arrives" is for the read stream, what about the write case here? Can we guarantee that a message will eventually be written out?

MadHed:
On my machine (Windows) the write callback is called all the time (`stream_select` never blocks). Which is actually a bit worrysome because without doing anything the php process soncumes about 14% cpu time.

MadHed:
I think the correct procedure would be to call `Loop\addWriteStream()` when `write()` is called and `Loop\removeWriteStream()` when the last message was compeltely sent.

MadHed:
The last commit must be handled with care! Somehow vscode completely hangs after parsing about 900 files and I have no idea why. It seems to be some kind of timing problem, because after adding logging calls the bug disappears.

src/ProtocolStreamWriter.php
```diff
@@ -26,22 +38,57 @@ public function __construct($output)
...
+     */
+    public function writeData()
+    {
+        $message = $this->messages[0]['message'];
+        $promise = $this->messages[0]['promise'];
+
+        error_clear_last();
+        $bytesWritten = @fwrite($this->output, $message);
+
+        if ($bytesWritten === false) {
```
kaloyan-raev: This clause would never be satisfied. If using `@`, `fwrite` will return 0 on error instead of FALSE. The whole error handling needs to be removed.

src/ProtocolStreamWriter.php
```diff
@@ -26,22 +38,57 @@ public function __construct($output)
...
+            array_shift($this->messages);
+
+            // This was the last message in the queue, remove the write handler.
+            if (count($this->messages) === 0) {
+                Loop\removeWriteStream($this->output);
+            }
+
+            $promise->fulfill();
+        } else {
+            $this->messages[0]['message'] = $message;
```
kaloyan-raev: What does happen with the promise in this case?

src/ProtocolStreamWriter.php
kaloyan-raev: I was not completely correct, but still this clause would be never satisfied.

`fwrite` returns FALSE only if invalid arguments are passed. `fwrite` will return 0 (or even higher value) on all other runtime errors, like "broken pipe" or "non-blocking operation not possible".

Please read this comment: http://php.net/manual/en/function.fwrite.php#96951

MadHed: Yeah, already forgot about this. Should we just ignore it then?

src/ProtocolStreamWriter.php
felixfbecker: > BTW, is there any part of the code that relies on these promises? I don't find any.

No, not yet because we don't have any functions that have async return values. The only async function is `indexProject()`, but it doesn't have a return value. I was planning to use more promises though, in particular here in the ProtocolStreamWriter and anything that involves file IO

src/ProtocolStreamWriter.php
felixfbecker: Yep, as it is not a runtime error.

kaloyan-raev:
I've done some testing of this PR. It works OK for me. I did not have any hangs.

If I add [this check](https://github.com/felixfbecker/php-language-server/blob/83afa0c1b8f779d339d2034e4e6fdc154762fdb3/src/ProtocolStreamReader.php#L33-L35), I also have the issue with orphan php processes resolved.

MadHed:
Removed the `false` check as per https://github.com/felixfbecker/php-language-server/pull/112#pullrequestreview-5589368

Also modified the writer to always try to write all pending messages in one tick.
The indexer now runs through without problems on Windows.

IMO another good addition would be to add a `writeUrgent()` which puts the message in front of the pending messages list. But this can be added later if needed. This would allow us to send unimportant messages like Diagnostics and LogMessage with a low priority while serving Symbol requests and the like immediately.

felixfbecker:
Why should it try to write all pending messages in one tick?

MadHed:
> Why should it try to write all pending messages in one tick?

The freeze I encountered had to do something with the amount of queued messages. Still not sure what exactly happened there. But since we are sending at least 2 messages per parsed, file the queue would grow constantly and eventually vscode window would freeze up. This fixes it.

kaloyan-raev:
My feeling is that this would make the language server more responsive during indexing. I like the idea for `writeUrgent()`.

MadHed:
I think writing all pending messages is correct for responsiveness reasons. If we were to write only a single message per tick the queue would quickly fill up while we are processing other server requests. Since we are not constrained by bandwidth this would artificially increase our response time. The writer still remains nonblocking because it yields as soon as a message could not be sent in full.

src/ProtocolStreamWriter.php
```diff
@@ -59,36 +59,33 @@ public function write(Message $msg)
...
-        $message = $this->messages[0]['message'];
-        $promise = $this->messages[0]['promise'];
+        $keepWriting = true;
+        while ($keepWriting) {
+            $message = $this->messages[0]['message'];
+            $promise = $this->messages[0]['promise'];
 
-        error_clear_last();
-        $bytesWritten = @fwrite($this->output, $message);
+            error_clear_last();
```
kaloyan-raev: Is this call still necessary?

felixfbecker:
Any reason why this much simpler version would not work;
```php
$data = (string)$msg;
        $msgSize = strlen($data);
        $writeData = function () use (&$data, $msgSize, &$writeData) {
            $bytesWritten = @fwrite($this->output, $data);
            if ($bytesWritten < $msgSize) {
                $data = substr($data, $bytesWritten);
            } else {
                Loop\removeWriteStream($this->output, $writeData);
            }
        };
        Loop\addWriteStream($this->output, $writeData);
```
?

MadHed:
@felixfbecker Do you mean for `write()`?
The Sabre event loop only allows a single callback per stream. Also the messages still have to be sent one after another, otherwise they would become mixed up.

felixfbecker:
I think I am fine with this. The only question left open is about the call sites - before a call would block, so now the methods that call `write()` need to return the promise too.

felixfbecker:
The only places though where this is used currently is the client, and there is makes more sense to actually return a promise for the response. Not sure how to handle that yet.

MadHed:
> The only places though where this is used currently is the client, and there is makes more sense to actually return a promise for the response. Not sure how to handle that yet.

Yeah, I think it would make more sense at a higher abstraction level where LSP requests and responses are handled asynchronously via promises.

PR #137, opened 2016-11-06
Log only to STDERR

### LINKED ISSUE #135: Only log exceptions to STDERR
Currently gets written to both STDOUT and STDERR which messes up with the protocol of course.

felixfbecker:
Closes #135 

Doesn't work currently:
```
PHP Fatal error:  Maximum function nesting level of '1000' reached, aborting! in C:\Users\felix\git\OpenSource\php-language-server\vendor\monolog\monolog\src\Monolog\Formatter\NormalizerFormatter.php on line 60
PHP Stack trace:
PHP   1. Symfony\Component\Debug\ErrorHandler->handleException() C:\Users\felix\git\OpenSource\php-language-server\vendor\symfony\debug\ErrorHandler.php:0
PHP   2. Monolog\Logger->log() C:\Users\felix\git\OpenSource\php-language-server\vendor\symfony\debug\ErrorHandler.php:528
PHP   3. Monolog\Logger->addRecord() C:\Users\felix\git\OpenSource\php-language-server\vendor\monolog\monolog\src\Monolog\Logger.php:517
PHP   4. Monolog\Handler\AbstractProcessingHandler->handle() C:\Users\felix\git\OpenSource\php-language-server\vendor\monolog\monolog\src\Monolog\Logger.php:336
PHP   5. Monolog\Formatter\LineFormatter->format() C:\Users\felix\git\OpenSource\php-language-server\vendor\monolog\monolog\src\Monolog\Handler\AbstractProcessingHandler.php:35
PHP   6. Monolog\Formatter\NormalizerFormatter->format() C:\Users\felix\git\OpenSource\php-language-server\vendor\monolog\monolog\src\Monolog\Formatter\LineFormatter.php:68
PHP   7. Monolog\Formatter\NormalizerFormatter->normalize() C:\Users\felix\git\OpenSource\php-language-server\vendor\monolog\monolog\src\Monolog\Formatter\NormalizerFormatter.php:43
PHP   8. Monolog\Formatter\NormalizerFormatter->normalize() C:\Users\felix\git\OpenSource\php-language-server\vendor\monolog\monolog\src\Monolog\Formatter\NormalizerFormatter.php:82
PHP   9. Monolog\Formatter\NormalizerFormatter->normalize() C:\Users\felix\git\OpenSource\php-language-server\vendor\monolog\monolog\src\Monolog\Formatter\NormalizerFormatter.php:82
PHP  10. Monolog\Formatter\NormalizerFormatter->normalize() C:\Users\felix\git\OpenSource\php-language-server\vendor\monolog\monolog\src\Monolog\Formatter\NormalizerFormatter.php:82
PHP  11. Monolog\Formatter\NormalizerFormatter->normalize() C:\Users\felix\git\OpenSource\php-language-server\vendor\monolog\monolog\src\Monolog\Formatter\NormalizerFormatter.php:82
PHP  12. Monolog\Formatter\NormalizerFormatter->normalize() C:\Users\felix\git\OpenSource\php-language-server\vendor\monolog\monolog\src\Monolog\Formatter\NormalizerFormatter.php:82
PHP  13. Monolog\Formatter\NormalizerFormatter->normalize() C:\Users\felix\git\OpenSource\php-language-server\vendor\monolog\monolog\src\Monolog\Formatter\NormalizerFormatter.php:82
```
Don't know why.

### DIFF
--- a/bin/php-language-server.php
+++ b/bin/php-language-server.php
@@ -3,6 +3,8 @@
 use LanguageServer\{LanguageServer, ProtocolStreamReader, ProtocolStreamWriter};
 use Sabre\Event\Loop;
 use Symfony\Component\Debug\ErrorHandler;
+use Monolog\Logger;
+use Monolog\Handler\StreamHandler;
 
 $options = getopt('', ['tcp::', 'memory-limit::']);
 
@@ -15,7 +17,11 @@ foreach ([__DIR__ . '/../../../autoload.php', __DIR__ . '/../autoload.php', __DI
     }
 }
 
-ErrorHandler::register();
+$logger = new Logger('Errors');
+$logger->pushHandler(new StreamHandler(STDERR));
+$errorHandler = new ErrorHandler;
+$errorHandler->setDefaultLogger($logger);
+ErrorHandler::register($errorHandler);
 
 @cli_set_process_title('PHP Language Server');
 
--- a/composer.json
+++ b/composer.json
@@ -29,7 +29,8 @@
         "sabre/event": "^5.0",
         "felixfbecker/advanced-json-rpc": "^2.0",
         "squizlabs/php_codesniffer" : "^2.7",
-        "symfony/debug": "^3.1"
+        "symfony/debug": "^3.1",
+        "monolog/monolog": "^1.21"
     },
     "minimum-stability": "dev",
     "prefer-stable": true,
--- a/tests/LanguageServerTest.php
+++ b/tests/LanguageServerTest.php
@@ -46,4 +46,19 @@ class LanguageServerTest extends TestCase
             ]
         ], $msg->body->result);
     }
+
+    public function testIndexing()
+    {
+        $input = new MockProtocolStream;
+        $output = new MockProtocolStream;
+        $output->on('message', function (Message $msg) {
+            var_dump($msg);
+        });
+        $server = new LanguageServer($input, $output);
+        $capabilities = new ClientCapabilities;
+        $capabilities->xcontent = true;
+        $capabilities->xglob = true;
+        $server->initialize(getmypid(), $capabilities, __DIR__);
+        \Sabre\Event\Loop\run();
+    }
 }


Seldaek:
The failure might be due to you passing a recursive array structure as monolog's context info. Looks like it comes from the exception handling though so I am not sure what exactly does it.. Do you have a minimal way to reproduce this so I could try to look what's up in monolog and possibly resolve this?

felixfbecker:
I just added these lines right after the register call:
```php
$arr = [];
echo $arr['undefined'];
```
causes an undefined index notice, which triggers the error handler.

What exactly do you mean with recursive array?

felixfbecker:
Help would be greatly appreciated

Seldaek:
I checked out the branch, added those two lines after ErrorHandler::register() in bin/php-language-server.php but I don't get any output when I run that file, and no error when I run phpunit... Not sure how you got this to fail?

Seldaek:
Ah nevermind sorry I forgot I had xdebug disabled in my 7.1RC6.. I can reproduce it on 7.0 - will investigate.

Seldaek:
OK that was really nasty. I pushed a fix to symfony to avoid the problem you hit (see ref above). That said, everything should work fine as long as you avoid errors on the top level.

Try doing this instead:

```
function testLog() {
    $arr=[];
    echo $arr['undefined'];
}
testLog();
```

Then it should not recurse endlessly anymore and you can check that it works.

felixfbecker:
Thanks a lot! Can you notify me when it is released?

Seldaek:
Well it'll be included in the next patch release, but to be honest as I said this only happens if an error occurs in the global scope which is highly unlikely in reality, so I'd say go ahead and merge this already, I wouldn't wait on a symfony release.

felixfbecker:
Okay - if my usage is otherwise correct, I will merge. Thanks!

felixfbecker:
Unfortunately, the actual issue is not solved - the errors are logged to STDERR, but the fatal error that is triggered by the uncaught exception is logged to STDOUT:

image

The output also is super ugly now as for some reason the exception becomes JSON-encoded instead of casted to a string with human-readable stack. Any idea? The symfony ExceptionHandler class does not have a `setDefaultLogger` method

PR #152, opened 2016-11-16
Tweak monolog config

Seldaek:
This changes the config a bit so I get something like this:

```
[2016-11-16 15:55:33] ERROR: Uncaught Exception ErrorException: "Undefined index: undefined" at .\php-language-server\bin\php-language-server.php line 54 {"exception":"[object] (ErrorException(code: 0): Undefined index: undefined at .\php-language-server\\bin\\php-language-server.php:54)\n[stacktrace]\n#0 .\php-language-server\\bin\\php-language-server.php(54): errorHandler(8, 'Undefined index...', '...', 54, Array)\n#1 .\php-language-server\\bin\\php-language-server.php(56): err()\n#2 {main}"}
```

It's still serialized in one line, that will be fixed in monolog 2.0 but not before unfortunately, but at least you don't get duplicate (or triple..) output and it is only sent to STDERR.

### DIFF
--- a/bin/php-language-server.php
+++ b/bin/php-language-server.php
@@ -2,9 +2,10 @@
 
 use LanguageServer\{LanguageServer, ProtocolStreamReader, ProtocolStreamWriter};
 use Sabre\Event\Loop;
-use Symfony\Component\Debug\ErrorHandler;
 use Monolog\Logger;
 use Monolog\Handler\StreamHandler;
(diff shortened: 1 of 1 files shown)


felixfbecker:
I ended up just not using any library at all, because with all the config I'm not saving any code in the end. Thanks for the effort though

PR #153, opened 2016-11-17
Fix formatting makes LS non-responsive

### LINKED ISSUE #144: Formatting makes LS non-responsive
If I call the Format action then:

1. The document is not formatted.
2. The language server becomes non-responsive. Nothing else seems to work anymore.

This seems to be a recent regression.

mniewrzal:
Closes #144

### DIFF
--- a/src/Formatter.php
+++ b/src/Formatter.php
@@ -8,11 +8,17 @@ use LanguageServer\Protocol\ {
     Range,
     Position
 };
-use PHP_CodeSniffer;
 use Exception;
+use PHP_CodeSniffer\ {
+    Config,
+    Ruleset,
+    Files\DummyFile,
+    Util\Tokens
+};
 
 abstract class Formatter
 {
+
     /**
      * Generate array of TextEdit changes for content formatting.
      *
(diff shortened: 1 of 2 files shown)


src/Formatter.php
```diff
@@ -8,11 +8,17 @@
     Range,
     Position
 };
-use PHP_CodeSniffer;
 use Exception;
+use PHP_CodeSniffer\ {
+    Config,
+    Ruleset,
+    Files\DummyFile,
+    Util\Tokens
```
mniewrzal: Sure.

src/Formatter.php
```diff
@@ -24,10 +30,24 @@
...
         $path = uriToPath($uri);
-        $cs = new PHP_CodeSniffer();
-        $cs->initStandard(self::findConfiguration($path));
-        $file = $cs->processFile(null, $content);
+        $config = new Config(['dummy'], false);
+        $config->standards = self::findConfiguration($path);
+
+        // Create this class so it is autoloaded and sets up a bunch
+        // of PHP_CodeSniffer-specific token type constants.
+        new Tokens();
```
mniewrzal: I will check this.

src/Formatter.php
```diff
@@ -42,9 +41,9 @@ public static function format(string $content, string $uri)
         $config = new Config(['dummy'], false);
         $config->standards = self::findConfiguration($path);
 
-        // Create this class so it is autoloaded and sets up a bunch
+        // Autoload class to sets up a bunch
         // of PHP_CodeSniffer-specific token type constants.
-        new Tokens();
+        spl_autoload_call('PHP_CodeSniffer\Util\Tokens');
```
felixfbecker: `Tokens::class` is nicer than string. Btw looked at the class, how can someone think it is a good idea to globally define constants in a class file. Total violation of PSR-2, a file should either define a symbol or cause side-effects. Ironic for a coding-style checker to violate the standard itself

PR #161, opened 2016-11-18
Add size limit back

### LINKED ISSUE #158: Size limit for indexed file is missing
I noticed that size limit for indexed file is missing. This is the reason for longer indexing and bigger memory consumption in my case.

felixfbecker:
Closes #158 
@mniewrzal Not sure about the size limit. Previously we had 0.5MB, but that includes files like the (generated) PhpParser files take about a second to parse alone and is 0.13MB / 2751 loc.
A 0.1MB size limit ignores files like phpunit/tests/Framework/AssertTest.php (0.12MB, 4500 loc) which could be considered "normal".

### DIFF
--- a/src/ContentTooLargeException.php
+++ b/src/ContentTooLargeException.php
@@ -1,0 +1,45 @@
+<?php
+declare(strict_types = 1);
+
+namespace LanguageServer;
+
+/**
+ * Thrown when the document content is not parsed because it exceeds the size limit
+ */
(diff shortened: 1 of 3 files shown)


mniewrzal:
For me you can leave it as it is in PR (0.15MB). It can be tuned later if there will be such need.

felixfbecker:
I will leave the decision to you since it was your feature request

Issue #54, opened 2016-10-09
Resolve types recursively


felixfbecker:
From #49:

Go To Definition currently works for
- class instantiations
- static method calls
- class constant access
- static property access
- parameter type hints
- return type hints
- method calls, if the variable was assigned to a new object in the same scope
- property access, if the variable was assigned to a new object in the same scope
- variables
- parameters
- imported closure variables (`use`)
- `use` statements that import a class-like
- class-like after `implements`/`extends`
- function calls
- constant access

What does not work yet:

``` php
$x = new Whatever();
$y = $x;
$y->myMethod(); // cannot get definition for myMethod

$x = new Whatever();
$x->property->myMethod(); // cannot get definition for myMethod
$x->myMethod()->anotherMethod(); // cannot get definition for anotherMethod

$this->myMethod(); // cannot get definition for myMethod, $this is not resolved to the class

trait MyTrait {}
class MyClass {
  use MyTrait; // cannot get definition for MyTrait
}
```
- Method/property access on a return value needs to be resolved _recursively_
- Variable reassignments need to be resolved _recursively_
- DocBlocks need to be parsed

z-ax:
Hi. Just checked with VSCode. I see a great work was done here but not yet looks usable to me.
Maybe, I'm doing something wrong, or that was not yet released to the VSCode extension?

From the checklist must say that `class constant access` doesn't work. As well as class constants auto-complete. I get huge list of everything except class constants, actually when typing `self::`

<img width="396" alt="screen shot 2016-11-01 at 01 17 00" src="https://cloud.githubusercontent.com/assets/1391015/19874866/017972e6-9fd1-11e6-96c3-cb4a004cda7d.png">

That list includes static functions, but not constants.

Also, can't get any completion for anything of other classes, even the list of available classes.

And no PhpDoc reading. E.g. I have a view with header like this:
```php
<?
/**
 * ...
 *
 * @var View $this
 * @var Meta $meta
 * @var bool $isUser
 * @var array|bool $release
 * @var array|bool $next
 * @var array|bool $previous
 */
```
No auto-complete at all as well.

felixfbecker:
@z-ax This issue is about _go to definition_, not completion. Completion is tracked by #9

z-ax:
Well, _go to definition_ doesn't work for those as well.
Thanks, will subscribe for both threads.

felixfbecker:
[...]

PR #165, opened 2016-11-19
Completion

### LINKED ISSUE #9: Completion
Should suggest:
- global symbols (classes, interfaces, traits, functions)
- local variables
- local parameters 
- variables imported with `use`
- not the variables outside of the function scope
- properties and methods of classes if the cursor is behind a `->`
  - it should resolve the variable to a class type and get the members
- namespaces and classes when the cursor is inside a `use` statement
- built-in PHP functions, classes and globals
  - how to get them? Is there a machine-parseble version of the docs?

It should:
- follow PSR autoloading rules defined in composer.json
- parse docblocks for type annotations of parameters and properties

felixfbecker:
Closes #9 💪
Big ups to @nikic for making this possible with the improvements in error recovery in v3!

animation

Works for properties and methods with all of the recursive type goodness of #54 

## Todo:
- [x] variables
- [x] keywords
- [x] classes
- [x] namespaces
- [x] constants
- [x] differentiate between static and instance properties/methods

## Testing
Two things are preventing a nice testing experience:
- VS Code has its own inferior completion. The only way to disable is going to the installation dir under resources/app/extensions/php/out/phpMain.js and commenting out the registration of the completionProvider.
- https://github.com/Microsoft/vscode-languageserver-node/issues/127 - I went inside the JS file and added a check for `null` (PR https://github.com/Microsoft/vscode-languageserver-node/pull/128)


mniewrzal:
Hi, I'm trying to play with this PR but I cannot make it working. I'm using code from gif. Is there anything specific I need to do to test it?

felixfbecker:
Are trying with VS Code? You need to manually patch your VS Code installation to fix the bugs in VS Code
Besides, this is WIP, maybe latest commit is broken. Are you getting any stack trace?

mniewrzal:
Yes, I'm using vscode 1.7.2, but without patching. Only stack trace is for `$a->`':
```
shell.ts:440 Cannot read property 'range' of null: TypeError: Cannot read property 'range' of null
    at asTextEdit (/home/wywrzal/git/vscode-php-intellisense/node_modules/vscode-languageclient/lib/protocolConverter.js:100:46)
    at /home/wywrzal/git/vscode-php-intellisense/node_modules/vscode-languageclient/lib/protocolConverter.js:95:67
    at set (/home/wywrzal/git/vscode-php-intellisense/node_modules/vscode-languageclient/lib/protocolConverter.js:83:13)
    at asCompletionItem (/home/wywrzal/git/vscode-php-intellisense/node_modules/vscode-languageclient/lib/protocolConverter.js:95:9)
    at Array.map (native)
    at asCompletionResult (/home/wywrzal/git/vscode-php-intellisense/node_modules/vscode-languageclient/lib/protocolConverter.js:76:26)
    at process._tickCallback (internal/process/next_tick.js:103:7)
```
If you feel its to early to test it let me know, I can wait.

felixfbecker:
That is the described bug in VS Code. I have proposed a fix here: https://github.com/Microsoft/vscode-languageserver-node/pull/128/files

You need to go into `~/.vscode/extensions/felixfbecker.vscode-php-intellisense-x.x.x/node_modules/vscode-languageclient/out/protocolConverter.js` and patch it there, then it works fine.

mniewrzal:
Ok, I will give it a try.

mniewrzal:
With patch completion works as described :) Thanks!

felixfbecker:
I think this is ready. Feedback? :)

kaloyan-raev:
Cool! I will test with Che on Monday.

________________________________

tests/LanguageServerTest.php
```diff
@@ -36,7 +36,10 @@ public function testInitialize()
                 'textDocumentSync' => TextDocumentSyncKind::FULL,
                 'documentSymbolProvider' => true,
                 'hoverProvider' => true,
-                'completionProvider' => null,
+                'completionProvider' => (object)[
+                    'resolveProvider' => false,
+                    'triggerCharacters' => ['$', '>']
```
kaloyan-raev: '>' should be '->', isn't it? I think we also need to add '::' here.

felixfbecker: Only characters are allowed in the array unfortunately. Using `->` will trigger the completion on `-` instead, as VS Code only looks at the first char. We should open an issue at LSP to support strings, this is a pretty big limitation for PHP and C++. `:` should be added

felixfbecker: That is because JSON does not have a "character" type, only a string type. But the property name clearly says trigger_Characters_, so if trigger _sequences_ are supported, it should be clarified in the protocol documentation.

kaloyan-raev: Obviously it can be read in different ways. So a clarification in the protocol would be helpful.

src/CompletionProvider.php
```diff
@@ -0,0 +1,383 @@
...
+        } else if ($node instanceof Node\Stmt\InlineHTML || $pos == new Position(0, 0)) {
+            $item = new CompletionItem('<?php', CompletionItemKind::KEYWORD);
+            $item->textEdit = new TextEdit(
+                new Range($pos, $pos),
+                stripStringOverlap($doc->getRange(new Range(new Position(0, 0), $pos)), '<?php')
+            );
+            $items[] = $item;
+        }
+
+        return $items;
```
kaloyan-raev: Please return `CompletionList` instead of `CompletionItem[]`. Due to https://github.com/Microsoft/language-server-protocol/issues/39 the Java binding supports only `CompletionList`.

felixfbecker: Okay, no problem. Is this the case for other places in the protocol too? How do you handle `MarkedString | string`? Pretty big limitation of the Java bindings if you ask me... Couldn't this be solved with polymorphism/generics?

kaloyan-raev: I haven't noticed any issue with the hovering feature.

You can see the arguments of the people behind the Java binding in the protocol's issue. I haven't dug into it. It seems there is some soft of agreement that the protocol will avoid mixed return types in future.

kaloyan-raev:
After fixing locally the issue with the `CompletionList` return type, I am able to test it with Eclipse Che, but I don't get the expected completion result on lots of places. Some examples:
- I get nothing here: https://github.com/felixfbecker/php-language-server/blob/master/bin/php-language-server.php#L7. In VSCode I see some results, but they don't come from the language server.
- Here: https://github.com/felixfbecker/php-language-server/blob/master/src/ClientHandler.php#L28, after "this->" I get only one proposal: "protocolReader". I should see also all properties and methods. I have no results if I just have "this->" on the line alone...
- I get the available variables only if I have typed "$". They should be included in the result even if I havent typed "$" yet.

felixfbecker:
Please see the tests for what is supported & tested: https://github.com/felixfbecker/php-language-server/tree/completion/fixtures/completion

> I get nothing here: https://github.com/felixfbecker/php-language-server/blob/master/bin/php-language-server.php#L7. In VSCode I see some results, but they don't come from the language server

Might be because it's in global scope, there is only a test for function scope.

> Here: https://github.com/felixfbecker/php-language-server/blob/master/src/ClientHandler.php#L28, after "this->" I get only one proposal: "protocolReader". I should see also all properties and methods. I have no results if I just have "this->" on the line alone...

If you request completion at the `PropertyFetch` node, it will be filtered by the property name that is already typed. So if `protocolReader` is already there, it will be the only suggestion.

>  I have no results if I just have "this->" on the line alone...

That might be the same issue as #171... The type is resolved to `Types\This`, but that of course needs to be resolved to the containing class. Maybe that's not done.

> I get the available variables only if I have typed "$". They should be included in the result even if I havent typed "$" yet.

Yes, variable suggestions are only provided when the node under the cursor is a `Variable` node.

Please note that a lot of features are not related to the completion support, but rather to the definition resolver (which means it also affects go-to-def and find-all-references).
I would be interested in whether the completion scenarios that are unit-tested (see fixtures/completion) work in Eclipse Che?

kaloyan-raev:
Another issue is that most of the time I don't see the `insertText` or the `textEdit` calculated properly.

Example 1. Type "ec" in new PHP file with only `<?php` tag. If I invoke code completion, I get only one proposal for "echo" with `insertText = "echo "`. The correct `insertText` in this case should be "ho ".

Example 2. Open https://github.com/felixfbecker/php-language-server/blob/master/bin/php-language-server.php#L6, place cursor right after "$op" and invoke code completion. I get a single result "$options" with 
```json
textEdit: { 
  "newText": "tions", 
  "range": {
    "start": {
      "line": 5,
      "character": 3,
    },
    "end": {
      "line": 5,
      "character": 3,
    }
  }
}
```
Leaving aside that "$options" should not be proposed here, the `newText` is correct, while the end of the range should not be equal to the start of the range. It should be correctly calculated to replace the current variable.

Note that the above examples have no issue with VSCode, because VSCode has a very sophisticated logic on the client side for applying completion proposals. Such logic is not described in LSP and my guess is that VSCode has it before the era of LSP. However, we should not expect other clients to have similiar logic implemented.

We should strive to have a `textEdit` with correct `newText` and `range` implemented for all `CompletionItem`s returns by the language service. This would guarantee the best integration for all clients.

felixfbecker:
Yes, that would be optimal. I started with just labels and moved some stuff already to using `textEdit`, with looking at the text before the cursor. Actually that logic should use the nodes more though and less the text content. I do not yet look at the text after the cursor, because that is not the common use case (TypeScript doesn't do this either).

kaloyan-raev:
> I would be interested in whether the completion scenarios that are unit-tested (see fixtures/completion) work in Eclipse Che?

Sure. I will go over the unit tests and try them in Che. Would you accept a PR to this "completion" branch with correction to the existing tests?

felixfbecker:
> Sure. I will go over the unit tests and try them in Che. Would you accept a PR to this "completion" branch with correction to the existing tests?

Of course. I don't know when I will have the time to do the other improvements so otherwise I was thinking about merging as-is and doing improvements later. But I can resolve the small things like returning `CompletionList` first.

kaloyan-raev:
OK. Then please fix the CompletionList and merge to master. We will invest time on our side to prepare PRs with the fixes for the issues I mentioned.

felixfbecker:
The issue I see with `CompletionList` is that returning `CompletionItem[]` lets the editor decide if it should re-request the completion list after further typing. But when returning `CompletionList`, you either tell the client that it should _definitely_ re-request, or _definitely not_.

kaloyan-raev:
> returning CompletionItem[] lets the editor decide if it should re-request the completion list after further typing.

I don't see anything positive in letting the editor decide on this. It is the server who really knows if the completion results are still good after further typing.

BTW, I think that for now we should set the `isIncomplete` to `true`. After fixing some of the issues, we can set it to `false`.

felixfbecker:
I thought about it, in theory the list of subsequent requests will only contain _less_ than the initial list. Am I right?

kaloyan-raev:
If the server sets the `isIncomplete` flag to `false` this means that further typing of the same "word" will result in returning the _same_ result. So, in this case it is pointless for the editor to set another request - it will get the same result. After applying the filtering rules, which is responsibility of the editor, the user will see _less_ results in the list.

felixfbecker:
In VS Code, if `isIncomplete` is `false`, the editor will filter the results client-side

kaloyan-raev:
As far as I studied VSCode some time ago, it always filters the results client-side. The `isIncomplete` flag determines whether to send a Completion Request on further typing.

felixfbecker:
Yeah, so if the user types `$` and we send back
- `$var`
- `$param`
when the user types an additional `v`, with `isIncomplete: true` it would send another request, while with `false` it would just use client side filtering on the previous result. The output would be the same, so we could save that request, right?

kaloyan-raev:
Exactly!

felixfbecker:
So why do you think we should set it to `true`?

kaloyan-raev:
In the above example, if you invoke code assist on an empty line, it will return a list without the two variables (this is one of the issues I reported). Then, if you type `$` and we have `isIncomplete: false`, the editor won't send another request to get them in the result list.

There are other issues like this. So, for now it's better to set `isIncomplete: true` to avoid unnecessary confusion. I also think it will be easier to find issues this way.

tests/LanguageServerTest.php
felixfbecker: https://github.com/Microsoft/language-server-protocol/issues/138

felixfbecker:
@mniewrzal Had to update CodeSniffer to latest master because of a bug, but now the formatter tests fail. Looks like the `Config` class has been removed?

mniewrzal:
Ok, I will take a look at this.

felixfbecker:
Never mind, fixed it. I referenced the wrong branch in composer.json. Sorry for the ping.

mniewrzal:
I found one issue, such code:
```php
<?php

class Test
{
    public function getAAA()
    {
    }
    public function getBBB()
    {
    }
}
$a = new Test();
$a->|getAAA();
```
If you will invoke completion in place of `|` you will see only `getAAA` but also `getBBB` should be proposed and should replace `getAAA` if applied.

mniewrzal:
Another issue:
```php
<?php
$a->
$a->|getAAA();
```
Invoke completion in place of `|` and you should see:
```
[Error - 10:52:03 AM] Request textDocument/completion failed.
  Message: strlen() expects parameter 1 to be string, null given
  Code: -32603 
```

felixfbecker:
@mniewrzal can you get the full stack trace? You need to change [this line](https://github.com/felixfbecker/php-language-server/blob/master/src/LanguageServer.php#L85) to `(string)$e`

@kaloyan-raev I set `isIncomplete` to `true`.

mniewrzal:
@felixfbecker I updated issue description.

felixfbecker:
Fixed, thanks!

mniewrzal:
Another thing that is missing (and its important imho) is completion for types. Now its hard to get completion in many places. Simple example:
```php
<?php
interface Example {
}
class Test implements Ex| {
}
```
From code I see that completion is limited to closed list of nodes. I think it would be great that if node isn't from handled list (or there is problem with parser) then result list should be as big as possible (types and keywords). It will be filtered by vscode after first letters.

felixfbecker:
`extends` / `implements` is a scenario that should be supported & tested. Should be quite easy with a check for `Node\Name` and parent for `Node\Stmt\Class` where currently only `Node\Expr\New_` and `Node\Expr\ConstFetch` are accepted. We can do these improvements in follow-up PRs though.

mniewrzal:
I understand that its doable but until new cases will be supported maybe it would be good approach to return 'everything' that workspace contains. It will also help with cases were parser will be not error tolerant. Of course it can be done later :)

mniewrzal:
Another thing that should be considered is completion filtering in server code. This PR is using substring but vscode supports filtering on client side and its 'fuzzy' filtering (e.g. in CSS `bc` will give `border-color`) which is more useful in my opinion. 

In general vscode built-in language servers like CSS doesn't do filtering on server side, results list is depends on code context not on prefix. This part was left for client (at least for now). 

If you feel that this is too much for this PR just merge it and I will open issues for things I mentioned in previous comments :)

mniewrzal:
Ok, one more thing that isn't enhancement but a bug ;) Example:
```php
<?php
class Test
{
    public function def(){}
    public function abc()
    {
        $this->|
    }
}
```
No completion after `->`. I need add one letter (a or d) to have proposals.

felixfbecker:
The thing is that fuzzy filtering has much worse performance. Remember that we might have to iterate over 100k definitions. Checking the key for a prefix is fast. Doing `strpos` is slower already. Fuzzy search would be even slower. We also don't do fuzzy search in symbol search.

Is your last bug related to `$this`? There are tests for property access without prefix.

mniewrzal:
> The thing is that fuzzy filtering has much worse performance. Remember that we might have to iterate over 100k definitions.

I understand, what I wanted to say is that there are many approaches:
- leave `strpos` as it is
- remove `strpos` filtering completely (client will filter response)
- apply `strpos` only to places where performance is endangered

> Is your last bug related to $this? There are tests for property access without prefix.

Yes, looks like that. I checked this:
```php
<?php
class Test
{
    public function adef(){
        $a = new Test();
        $a->
    }
}
```
and works ok.

felixfbecker:
Yes, but sending 100k suggestions has a performance impact too. I think the current solution is the best compromise for now.

This might be #171 then

kaloyan-raev:
@felixfbecker I suggest that we follow the pattern already used for the VSCode language servers implemented by Microsoft, which is:
- Do context-based filtering, i.e. avoid returning completion item kinds that are not suited for the current cursor position.
- Don't do input-based filtering and leave this to the client. By "input-based filtering" I mean filtering based on the current word entered by the user.
- Return complete result lists, i.e. with `isIncomplete: false`

The pros of the above approach is:
- Client can decide on the input-based filtering algorithm. The fuzzy filtering is a great usability feature and we should give the opportunity to clients to implement it.
- Any additional typing after the initial completion request is very fast and does not involve further client-server communication.

Cons:
- The initial completion request may take some time due to the big completion result, e.g. tens of thousands of items. This is partially mitigated by the context-based filtering and especially by the faster updates on the additional typing.

If we decide on a stricter server-based filtering to reduce the size of the completion list then:
- We give up the fuzzy filtering, which is a pitty
- We give up on returning complete result lists, forcing the client to send completion requests on every keystroke on further typing after the initial completion request. This has its own performance implications.

felixfbecker:
What does "context-based" filtering mean? When giving completions for a method for example, we have to filter by an FQN prefix to filter by the class - so why not include the part of the typed property name too?

kaloyan-raev:
Example for "context-based" filtering. If you have `$this->meth|`, the completion list should return all visible methods and members of the current class (and its hierarchy), without filtering it by the "meth" input (the input-based filtering). Perhaps, variables should be included too to make something like `$this->$my_local_var" possible.

But the completion list should not include:
- non-visible methods and members, e.g. private methods of a superclass
- keywords
- global functions
- Class/Interface/Traits/Namespaces names
- etc.

felixfbecker:
yeah, gotcha. Just saying that the class-filtering is implemented atm by doing a FQN prefix search. So I don't really see the disadvantage of also excluding anything that doesn't match what is already typed? Besides fuzzy searching of course. It saves a few items being sent to the client that don't match anyway.

kaloyan-raev:
I wouldn't worry about the size of the completion list at that place. I don't believe a class can have thousands of methods and members.

The largest list is produced when you invoke code completion on an empty line.

Issue #173, opened 2016-11-21
Use language-server programmatically?


bonndan:
Hi Felix,

I am currently working on a PHP2Groovy transpiler which uses quite the same libs as your language server, and of course I run into the same challenges (type inference, docblock parsing...). Now my idea is to get rid of some of my code and have the language server do some of the code inspection. However, I'd like to use it programmatically from within my PHP process. How is that possible?

And could you drop me 1-2 lines explaining what kind of type inspection you do inside functions/methods and between classes?

Viele Grüße aus Bonn

felixfbecker:
I did think about building the LS in two parts: one package for the protocol interface, one package for the intelligence that can be used programmatically. I decided against it because the LSP _is_ already a generic, programmatically usable interface. If I structured this as a library I would limit myself in a few ways. For example, I would not want to force consumers to use sabre's event loop, and things like spawning new processes for parsing would become more challenging. Currently the protocol is the only interface the server is meant to be consumed, and if you use it as a library, you might get breaking changes in between minor versions because I refactor classes etc. So what you _could_ do is spawn the server, send him some requests and use the answers. My JSON RPC dispatcher is in a separate package so you could use that.

As I think about it, I could of course at least make the `LanguageServer` class that resembles all the methods "public" API and refactor the protocol parser into a separate class. Then you could instantiate it and call the methods directly without the overhead and added complexity of IPC. You would still have to use the event loop though because everything is written for concurrency.

I don't know though if that would be particularly useful for a transpiler. The LSP is very oriented around IDE features, so what you get is always what would be shown in a UI to a user. You would probably be more interested in the internal classes, like `DefinitionResolver` - but event the `Definition` class is coupled to LSP structures. So there would be refactoring needed to decouple them all.

I am not against this at all, if you have an idea how to that the best way and wanna propose that / do a PR, I would be happy to merge it.

As for the type inference: Inside `DefinitionResolver` there is a method called `resolveExpressionNodeToType`. What it does is it takes any `PhpParser\Node\Expr` node and recursively tries to get the type that expression will resolve to at runtime. It will use `@var` tags of (indexed)  properties, method/function return types, find the place where a variable was assigned, convert scalars to a type. It even goes as far as a ternary will become a compound type of its expressions, and array will get properly inferred key and value types etc. I actually use the LS itself while developing the LS (eat your own dogfood), and everyday I find new ways where a type could be inferred and directly implement them (or at make myself a ticket).

[...]

PR #181, opened 2016-11-23
extracted a document factory

bonndan:
Related to https://github.com/felixfbecker/php-language-server/issues/173

may just have a look first to discuss things, don't necessarily merge

### DIFF
--- a/src/PhpDocumentFactory.php
+++ b/src/PhpDocumentFactory.php
@@ -1,0 +1,215 @@
+<?php
+declare(strict_types = 1);
+
+namespace LanguageServer;
+
+use LanguageServer\NodeVisitor\ColumnCalculator;
+use LanguageServer\NodeVisitor\DefinitionCollector;
+use LanguageServer\NodeVisitor\DocBlockParser;
+use LanguageServer\NodeVisitor\ReferencesAdder;
+use LanguageServer\NodeVisitor\ReferencesCollector;
+use LanguageServer\Protocol\ClientCapabilities;
+use LanguageServer\Protocol\Diagnostic;
+use LanguageServer\Protocol\DiagnosticSeverity;
+use phpDocumentor\Reflection\DocBlockFactory;
+use PhpParser\ErrorHandler\Collecting;
+use PhpParser\NodeTraverser;
+use PhpParser\NodeVisitor\NameResolver;
+use Sabre\Event\Promise;
+use function Sabre\Event\coroutine;
+
+class PhpDocumentFactory
+{
+    /**
+     * Holds account of all definitions, references etc.
+     * @var Project
+     */
+    private $project;
+    
+    /**
+     * Instance of the PHP parser
+     *
+     * @var Parser
+     */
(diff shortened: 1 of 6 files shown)


src/PhpDocumentFactory.php
```diff
@@ -0,0 +1,215 @@
...
+     * The client's capabilities
+     *
+     * @var ClientCapabilities
+     */
+    private $clientCapabilities;
+
+    public function __construct(LanguageClient $client, Project $project, ClientCapabilities $clientCapabilities = null)
+    {
+        $this->client = $client;
+        $this->clientCapabilities = $clientCapabilities ? : new ClientCapabilities();
```
felixfbecker: The document(factory)  should not know about clients or LSP. The content retrieval should happen over an injected interface like `ContentRetriever` then

src/PhpDocumentFactory.php
```diff
@@ -0,0 +1,215 @@
...
+                }
+                $content = file_get_contents($path);
+            }
+            
+            return $this->createDocument($uri, $content, false);
+        });
+    }
+
+    /**
+     * Ensures a document is loaded and added to the list of open documents.
```
felixfbecker: I find it a bit weird that a `createDocument` method of a factory causes side-effects. I would assume it to only create a document.

src/Project.php
```diff
@@ -354,4 +254,96 @@ public function isDefined(string $fqn): bool
...
+    }
+
+    /**
+     * Ensures a document is loaded and added to the list of open documents.
+     *
+     * @param string $uri
+     * @param string $content
+     * @return void
+     */
+    public function openDocument(string $uri, string $content)
```
felixfbecker: This would then call `addDocument`

src/Project.php
```diff
@@ -89,86 +61,14 @@ public function getDocument(string $uri)
...
-    /**
-     * Ensures a document is loaded and added to the list of open documents.
-     *
-     * @param string $uri
-     * @param string $content
+     * @param PhpDocument $document
      * @return void
      */
-    public function openDocument(string $uri, string $content)
+    public function addDocument(PhpDocument $document)
```
felixfbecker: This method could register all the definitions of the document to the index, instead of the factory. Would be one more iteration though.

src/Server/TextDocument.php
```diff
@@ -98,7 +98,7 @@ public function didOpen(TextDocumentItem $textDocument)
      */
     public function didChange(VersionedTextDocumentIdentifier $textDocument, array $contentChanges)
     {
-        $this->project->getDocument($textDocument->uri)->updateContent($contentChanges[0]->text);
+        $this->project->openDocument($textDocument->uri, $contentChanges[0]->text);
```
felixfbecker: You are messing a bit here with the semantics. `didChange` is not supposed to _open_ a document, it is supposed to _update_ an opened document.

bonndan:
Thanks for the feedback. I close this PR now and will open a different one.

PR #182, opened 2016-11-24
refactored project and phpdocument

bonndan:
After the last PR I noticed that the Project is more or less the index, extracting an Index class leads to the project being an empty shell. Same applied to the PHPDocumentFactory after the dependency code smell was removed.
For my usecase I'd handle with the statements in PhpDocuments mostly and with the Project. After all it probably makes no difference if this refactoring is in. So feel free to turn the PR down. I will continue to try integrating PLS into my project any maybe come up with an new insight.

### DIFF
--- a/src/PhpDocument.php
+++ b/src/PhpDocument.php
@@ -3,61 +3,15 @@ declare(strict_types = 1);
 
 namespace LanguageServer;
 
-use LanguageServer\Protocol\{Diagnostic, DiagnosticSeverity, Range, Position, TextEdit};
+use LanguageServer\Protocol\{Diagnostic, DiagnosticSeverity, Range, Position};
 use LanguageServer\NodeVisitor\{
-    NodeAtPositionFinder,
-    ReferencesAdder,
-    DocBlockParser,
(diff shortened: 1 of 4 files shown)


felixfbecker:
Separating the index into another class is great because I thought about one project having multiple indexes - one for the project itself, and one for every dependency. That way I can cache the dependencies individually by the installed version and across projects.
I still don't see the real benefit of this refactor, I feel you are trying to make the `PhpDocument` do less, but I don't see what's the benefit of that. I'm open for decoupling and refactoring, but I just don't see the use case here, so closing this one. If you meet any actual limitation of the current structure, feel free to do another PR.

Please note that as it stands I still follow semver only for the LSP interface. If you use the internal classes, stuff might break between minors. I would be open to separating the classes into a new package that is versioned independently so the LS can still provide soft versioning to consumers of the LSP.

felixfbecker:
FYI I split `Project` into `PhpDocumentLoader` and `Index` classes, and removed the dependency on the `LanguageClient` from both `PhpDocumentLoader` and `PhpDocument`.

PR #189, opened 2016-11-30
Display warning message about XDebug enabled

### LINKED ISSUE #188: Display warning message about XDebug enabled
Maybe it would be good idea to display info about XDebug enabled (like Composer). Request `window/showMessage` with MessageType::Warning can be used for example.

mniewrzal:
Closes #188


src/LanguageServer.php
```diff
@@ -125,6 +125,12 @@ public function initialize(ClientCapabilities $capabilities, string $rootPath =
             $this->indexProject()->otherwise('\\LanguageServer\\crash');
         }
 
+        if (extension_loaded('xdebug')) {
+            setTimeout(function () {
```
mniewrzal: I'm not sure if its best way to do this. This is added because tests are failing (notification message is first before initialization result).

felixfbecker:
Closing since the LS can now auto-restart without XDebug if XDebug is enabled

PR #191, opened 2016-11-30
Shutdown when the socket is closed

### LINKED ISSUE #190: Terminate server if stream was closed
If the socket was closed, the child process needs to exit

felixfbecker:
Closes #190

### DIFF
--- a/src/LanguageServer.php
+++ b/src/LanguageServer.php
@@ -65,6 +65,10 @@ class LanguageServer extends AdvancedJsonRpc\Dispatcher
     {
         parent::__construct($this, '/');
         $this->protocolReader = $reader;
+        $this->protocolReader->on('close', function () {
+            $this->shutdown();
+            $this->exit();
+        });
         $this->protocolReader->on('message', function (Message $msg) {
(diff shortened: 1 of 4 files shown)


src/LanguageServer.php
```diff
@@ -65,6 +65,10 @@ public function __construct(ProtocolReader $reader, ProtocolWriter $writer)
     {
         parent::__construct($this, '/');
         $this->protocolReader = $reader;
+        $this->protocolReader->on('close', function () {
+            $this->shutdown();
+            $this->exit();
```
keegancsmith: is shutdown and exit idempotent? ie is there any issue in the case of:

1. client sends shutdown, we run `$this->shutdown()`
2. client sends exit, we run `$this->exit()`
3. client closes TCP connection
4. server close event causes us to then call `$this->shutdown()` and `$this->exit()` again.

felixfbecker: At the moment, `shutdown` just frees the index to be garbage collected and `exit` will exit the process, so after that nothing can happen anymore

PR #207, opened 2016-12-07
Not all class members are proposed in completion

### LINKED ISSUE #194: Not all class members are proposed in completion
If statement with class members (e.g. method or field) is finished then invoking completion inside such element will not return all possible members with prefix before cursor. Example:
```php
$a = new TestClass();
$a->test|Method();
```
[TestClass](https://github.com/felixfbecker/php-language-server/blob/master/fixtures/global_symbols.php#L21) is from fixtures. If completion will be triggered in place of `|` then only `testMethod()` is proposed. Property `testProperty` should be also visible.

mniewrzal:
Closes #194

### DIFF
--- a/src/CompletionProvider.php
+++ b/src/CompletionProvider.php
@@ -4,12 +4,10 @@ declare(strict_types = 1);
 namespace LanguageServer;
 
 use PhpParser\Node;
-use phpDocumentor\Reflection\Types;
 use LanguageServer\Protocol\{
     TextEdit,
     Range,
     Position,
-    SymbolKind,
     CompletionList,
     CompletionItem,
     CompletionItemKind
@@ -156,8 +154,12 @@ class CompletionProvider
                     }
                 }
             } else {
-                $fqn = $this->definitionResolver->resolveReferenceNodeToFqn($node);
-                $prefixes = $fqn !== null ? [$fqn] : [];
+                $prefix = $this->definitionResolver->resolveReferenceNodeToFqn($node);
+                $index = ($index = strpos($prefix, '->')) ? $index : strpos($prefix, '::');
+                if ($index) {
+                    $prefix = substr($prefix, 0, $index + 2);
+                }
+                $prefixes = $prefix !== null ? [$prefix] : [];
             }
 
             foreach ($this->project->getDefinitions() as $fqn => $def) {
--- a/src/CompletionProvider.php
+++ b/src/CompletionProvider.php
@@ -132,34 +132,25 @@ class CompletionProvider
             || $node instanceof Node\Expr\StaticPropertyFetch
             || $node instanceof Node\Expr\ClassConstFetch
         ) {
-            if (!is_string($node->name)) {
-                // If the name is an Error node, just filter by the class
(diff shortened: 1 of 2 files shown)


src/CompletionProvider.php
```diff
@@ -156,8 +154,12 @@ public function provideCompletion(PhpDocument $doc, Position $pos): CompletionLi
                     }
                 }
             } else {
-                $fqn = $this->definitionResolver->resolveReferenceNodeToFqn($node);
-                $prefixes = $fqn !== null ? [$fqn] : [];
+                $prefix = $this->definitionResolver->resolveReferenceNodeToFqn($node);
+                $index = ($index = strpos($prefix, '->')) ? $index : strpos($prefix, '::');
```
felixfbecker: `$prefix` can be `null`, in which case the call to `strpos` is incorrect. This is why I saved it in the variable `$fqn`.
Also I find this very hard to read, I would just do `preg_split('/->|::/', $fqn)[0]`.
But tbh I think if you don't want to handle the case of already having typed a prefix, the check at line 137|135 can just go away and this `else` branch is not needed - it exists only to filter by an already-typed property name.

mniewrzal: Good point. Thanks for hint!

src/CompletionProvider.php
```diff
@@ -134,30 +132,25 @@ public function provideCompletion(PhpDocument $doc, Position $pos): CompletionLi
...
-            if (!is_string($node->name)) {
-                // If the name is an Error node, just filter by the class
+            // If the name is an Error node, just filter by the class
+            if ($node instanceof Node\Expr\MethodCall || $node instanceof Node\Expr\PropertyFetch) {
+                // For instances, resolve the variable type
+                $prefixes = DefinitionResolver::getFqnsFromType(
+                    $this->definitionResolver->resolveExpressionNodeToType($node->var)
+                );
+            } else {
+                $prefixes = [$node->class instanceof \PhpParser\Node\Name ? (string)$node->class : ''];
```
felixfbecker: Why this change? `Node` is used as a shorthand everywhere in the codebase

mniewrzal: For some reason previous condition returns false for case like 'TestClass::$st'. Most probably because there is namespace \PhpParser\Node\Name and there is class 'Name' in namespace `\PhpParser\Node`. I change it and I wanted to ask what will be the best solution for this :) You were faster.

felixfbecker: PhpParser does this pattern a lot - defining an abstract base class like `PhpParser\Node\Expr` and then several implementations in a namespace like `PhpParser\Node\Expr\Variable`. This works fine in all other cases. Could you check the actual value of `$node->class` in your error case?

mniewrzal: `$node->class` is `PhpParser\Node\Name\FullyQualified`. `(string)$node->class` gives proper class name.

felixfbecker: All tests pass when using `Node\Name`

mniewrzal: Hmm strange... but good that its not a problem in general. Thanks for help.

PR #247, opened 2017-01-19
Fix $this autocompletion

### LINKED ISSUE #195: No member completion for empty prefix
No completion for two separate cases.
```php
<?php
class TestClass 
{
    function abc(){
        $this->
    }
}
```
```php
TestClass::
```
If at least one letter will be added then completion will work.

vakata:
Closes #195 

There probably is a more elegant way to do this, but this is what I came up with a quick view of the available methods. Let me know if you need me to change anything (if the PR is acceptable at all of course :)).

### DIFF
--- a/src/CompletionProvider.php
+++ b/src/CompletionProvider.php
@@ -130,10 +130,6 @@ class CompletionProvider
         $list = new CompletionList;
         $list->isIncomplete = true;
 
-        //echo get_class($node->var);
-        //var_dump((string)$node->var->name);
-        //die();
-
         // A non-free node means we do NOT suggest global symbols
         if (
             $node instanceof Node\Expr\MethodCall
@@ -145,7 +141,7 @@ class CompletionProvider
             // If the name is an Error node, just filter by the class
             if ($node instanceof Node\Expr\MethodCall || $node instanceof Node\Expr\PropertyFetch) {
                 // For instances, resolve the variable type
-                 $prefixes = DefinitionResolver::getFqnsFromType(
+                $prefixes = DefinitionResolver::getFqnsFromType(
                     $this->definitionResolver->resolveExpressionNodeToType($node->var)
                 );
             } else {
--- a/src/CompletionProvider.php
+++ b/src/CompletionProvider.php
@@ -130,6 +130,10 @@ class CompletionProvider
         $list = new CompletionList;
         $list->isIncomplete = true;
 
+        //echo get_class($node->var);
+        //var_dump((string)$node->var->name);
+        //die();
+
         // A non-free node means we do NOT suggest global symbols
         if (
             $node instanceof Node\Expr\MethodCall
@@ -141,7 +145,7 @@ class CompletionProvider
             // If the name is an Error node, just filter by the class
             if ($node instanceof Node\Expr\MethodCall || $node instanceof Node\Expr\PropertyFetch) {
                 // For instances, resolve the variable type
-                $prefixes = DefinitionResolver::getFqnsFromType(
+                 $prefixes = DefinitionResolver::getFqnsFromType(
                     $this->definitionResolver->resolveExpressionNodeToType($node->var)
                 );
             } else {
--- a/src/CompletionProvider.php
+++ b/src/CompletionProvider.php
@@ -140,18 +140,9 @@ class CompletionProvider
         ) {
             // If the name is an Error node, just filter by the class
             if ($node instanceof Node\Expr\MethodCall || $node instanceof Node\Expr\PropertyFetch) {
-                // For instances, resolve the variable type
-                $classNode = null;
-                if ($node->var instanceof Node\Expr\Variable && $node->var->name === 'this') {
-                    $classNode = getClosestNode($node, Node\Stmt\Class_::class);
-                }
-                if ($classNode !== null && !$classNode->isAnonymous()) {
-                    $prefixes = [(string)$classNode->namespacedName];
-                } else {
-                    $prefixes = DefinitionResolver::getFqnsFromType(
-                        $this->definitionResolver->resolveExpressionNodeToType($node->var)
-                    );
-                }
+                $prefixes = DefinitionResolver::getFqnsFromType(
+                    $this->definitionResolver->resolveExpressionNodeToType($node->var)
+                );
             } else {
                 // Static member reference
                 $prefixes = [$node->class instanceof Node\Name ? (string)$node->class : ''];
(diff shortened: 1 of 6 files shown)


felixfbecker:
This should be done in `DefinitionResolver::resolveExpressionNodeToType()`.
And iirc there should be already code in place for this, apparently it is not working.
Appreciate you looking into this - even a failing unit test is already valuable.

vakata:
Thank you for the tip - sorry, I am still getting familiar with the code. I was worried that not returning the special `This` type might negatively affect other parts of the code, not related to autocompletion.

vakata:
@felixfbecker please let me know if there is anything else I can do to help further with this PR

felixfbecker:
> I was worried that not returning the special This type might negatively affect other parts of the code

This was the reason I had to think about this a bit more. `Types\This` is the correct type, because it can be a subclass of the containing class. Fluent interfaces should be supported, and I am not sure whether that will work this way:

```php
class A {
  public function a() {
    return $this;
  }
}
class B extends A {
  public function b() {
  }
}

$b = new B;
$b->a()->|
```
We would want suggestions for `a()` _and_ `b()` here, the return type should be `B`, not `A`.

Maybe you could add a test for this and check if it works?

Of course your solution is in any way better than the way it is at the moment.

vakata:
Currently the type is resolved correctly as `\B`, but the lookup for the `B->a()` definition fails. I believe this has always been this way - it is not a regression. 
I am trying to figure out how to check for inherited methods (in order to make this work), my guess is it will have something to do with getting the type of the instance, checking the `extends` property and checking all classes for definitions of the missing method. Am I on the right track? 
Which leads me to another question, are trait methods used with a different name currently handled?

vakata:
OK, this introduces code duplication, but I just wanted to see it will work. I took the `expandParentFqns` method and copied it to `definitionResolver` where I could now check all parent classes for the signature. I will see if I can improve on this tomorrow by moving the calls around to avoid duplication.

vakata:
The `expandParentFqns` addition was a bad call - CPU usage sky rocketed - will debug further.

felixfbecker:
This is probably very complicated to solve. I appreciate your effort - and would be okay with merging your initial solution until you finish your support for fluent interfaces, since it was in any case better than the current behaviour.

vakata:
I created another branch with fluent interfaces support, which uses a generator instead of collecting parents in an array: https://github.com/vakata/php-language-server/commit/7b4d0d1c6094c3cd760477a111dbf66c4786c297

This definitely had a positive effect, but I still feel it is a bit more sluggish than the version without the generator, it is a bit subjective though - I will have to create a proper benchmark to be sure.

src/DefinitionResolver.php
```diff
@@ -479,8 +483,10 @@ public function resolveExpressionNodeToType(Node\Expr $expr): Type
                 if ($expr instanceof Node\Expr\MethodCall) {
                     $fqn .= '()';
                 }
-                $def = $this->index->getDefinition($fqn);
-                if ($def !== null) {
+                if ($def = $this->index->getDefinition($fqn)) {
+                    if ($def->type instanceof Types\This || $def->type instanceof Types\Self_) {
+                        return $this->resolveExpressionNodeToType($expr->var);
+                    }
```
felixfbecker: Why is a check for `Types\This` needed when you now change the behaviour above to always resolve to `Types\Object` with the containing class?

What issue does the check for `Types\Self` fix? I only see a test for `$this`

src/CompletionProvider.php
```diff
@@ -141,7 +145,7 @@ public function provideCompletion(PhpDocument $doc, Position $pos): CompletionLi
             // If the name is an Error node, just filter by the class
             if ($node instanceof Node\Expr\MethodCall || $node instanceof Node\Expr\PropertyFetch) {
                 // For instances, resolve the variable type
-                $prefixes = DefinitionResolver::getFqnsFromType(
+                 $prefixes = DefinitionResolver::getFqnsFromType(
```
felixfbecker: Revert the changes in this file

src/DefinitionResolver.php
vakata: Because the `$this` reference may not come from a variable. 
If the check is removed the following example no longer works:
```php
<?php

class ThisClassPrefix extends TestClass
{
    private $foo;
    private $bar;

    /**
     * @return self
     */
    protected function method()
    {
        return $this;
    }
    public function test()
    {
        $this->method()->|
    }
}
```

felixfbecker: Your example answers my question about `Types\Self`, but it doesn't answer my first question (btw, the correct return type for `method()` would be `@return $this`)

vakata: Well, unfortunately I could not find a standard for this, and it seems `@return self` is very wide-spread. Maybe because it is the only option defined in the [phpDocumentor site](https://www.phpdoc.org/docs/latest/references/phpdoc/types.html)

As for `this` - I thought I was answering both questions - if you replace `@return self` with `@return $this` in the above example `$def->type` will be an instance of `phpDocumentor\Reflection\Types\This`. I believe this is the correct behavior and will be very important for fluent interfaces - all definitions in the index still use `Types\This`, it is evaluated as needed - hence the check. Maybe I am missing something?

felixfbecker: It is documented here: https://www.phpdoc.org/docs/latest/guides/types.html#keywords

Yes, it is needed for fluent interfaces, I just thought you would like to merge in a solution that fixes $this autocompletion without support for fluent interfaces first and then do a follow-up

vakata: Oh, sorry, I missed the `$this` keyword (I will have to rewrite a whole lot of stuff now :)) This is your call then - should I remove the check for `Self_`? Will there be any gain in removing it? I see only the downside of not supporting `return self` which even incorrectly is used a lot for indicating returnig `$this`. As I said - your call? 
As for `this` - I believe it is better this way - no point in breaking something that works. As said in the discussion earlier - some stuff (your example) is still not supported - so support is incomplete, but it will be a good foundation.

vakata: @felixfbecker should I remove the Self_ check? Or do you need more time to think about it?

src/DefinitionResolver.php
```diff
@@ -479,8 +483,10 @@ public function resolveExpressionNodeToType(Node\Expr $expr): Type
                 if ($expr instanceof Node\Expr\MethodCall) {
                     $fqn .= '()';
                 }
-                $def = $this->index->getDefinition($fqn);
-                if ($def !== null) {
+                if ($def = $this->index->getDefinition($fqn)) {
```
vakata: This is my opinion as well - I only adopted this style as I saw it used throughout the code. I will revert asap.

felixfbecker: There are places where it is only possible like this

jens1o:
any news on this?

vakata:
I will create a separate simpler pull request now that tolerant parses has been adopted.

PR #250, opened 2017-01-20
Signature help

### LINKED ISSUE #18: Signature help
https://github.com/Microsoft/language-server-protocol/blob/master/protocol.md#signature-help-request

image

vakata:
Closes #18
Parameter counting is "hackish" and working with php://temp maybe undesired? I also had to store parsed parameters along with the definition (similar to `declarationLine`) to avoid parsing again.


src/Definition.php
```diff
@@ -89,4 +89,11 @@ class Definition
      * @var string
      */
     public $documentation;
+
+    /**
+     * Parameters array (for methods and functions), for use in textDocument/signatureHelp
+     *
+     * @var string[]
```
felixfbecker: This should be `ParameterInformation[]`

src/DefinitionResolver.php
```diff
@@ -144,6 +150,10 @@ public function resolveReferenceNodeToDefinition(Node $node)
     {
         // Variables are not indexed globally, as they stay in the file scope anyway
         if ($node instanceof Node\Expr\Variable) {
+            // Resolve $this
+            if ($node->name === 'this' && $fqn = $this->getContainingClassFqn($node)) {
+                return $this->index->getDefinition($fqn, false);
+            }
```
felixfbecker: Please keep this out of this PR

src/LanguageServer.php
```diff
@@ -247,6 +248,9 @@ public function initialize(ClientCapabilities $capabilities, string $rootPath =
             $serverCapabilities->completionProvider = new CompletionOptions;
             $serverCapabilities->completionProvider->resolveProvider = false;
             $serverCapabilities->completionProvider->triggerCharacters = ['$', '>'];
+            // Support "Signature Help"
+            $serverCapabilities->signatureHelpProvider = new SignatureHelpOptions;
+            $serverCapabilities->signatureHelpProvider->triggerCharacters = ['('];
```
felixfbecker: Add comma as a trigger character

src/SignatureHelpProvider.php
```diff
@@ -0,0 +1,211 @@
...
+
+    /**
+     * @param DefinitionResolver $definitionResolver
+     * @param ReadableIndex      $index
+     */
+    public function __construct(DefinitionResolver $definitionResolver, ReadableIndex $index)
+    {
+        $this->definitionResolver = $definitionResolver;
+        $this->index = $index;
+        $this->parser = new Parser;
```
felixfbecker: Why do we need a new parser instance here?

src/SignatureHelpProvider.php
```diff
@@ -0,0 +1,211 @@
...
+     * Returns signature help for a specific cursor position in a document
+     *
+     * @param PhpDocument $doc The opened document
+     * @param Position $pos The cursor position
+     * @return SignatureHelp
+     */
+    public function provideSignature(PhpDocument $doc, Position $pos) : SignatureHelp
+    {
+        $help = new SignatureHelp;
+        $help->signatures = [];
```
felixfbecker: Change the constructor of `SignatureHelp` to initialize this

src/SignatureHelpProvider.php
```diff
@@ -0,0 +1,211 @@
...
+
+        $handle = fopen('php://temp', 'r+');
+        fwrite($handle, $doc->getContent());
+        fseek($handle, 0);
```
felixfbecker: We have the whole content as a string, so opening a stream for this provides no benefit at all.
Did you try solving this just with the PhpParser node offsets you get from `getNodeAtPosition()`?

tests/Server/TextDocument/SignatureHelpTest.php
```diff
@@ -0,0 +1,222 @@
...
+        $completionUri = pathToUri(__DIR__ . '/../../../fixtures/signatureHelp/funcClosed.php');
+        $this->loader->open($completionUri, file_get_contents($completionUri));
+        $result = $this->textDocument->signatureHelp(
+            new TextDocumentIdentifier($completionUri),
+            new Position(6, 10)
+        )->wait();
+
+        $help = new SignatureHelp;
+        $help->signatures = [];
+        $info = new SignatureInformation;
```
felixfbecker: Add constructors to the classes that allow to pass all properties as arguments so we can define the result inline without saving it to a variable

src/Definition.php
vakata: You want me to build the objects here? Won't that use more memory than plain strings, which are converted to ParameterInformation when needed?

src/SignatureHelpProvider.php
vakata: Sorry - that was leftover - I eventually used `Lexer` directly - will remove.

src/Definition.php
felixfbecker: If I understand your code correctly this string array only contains the pretty-printed parameter node. That means you cannot construct the `documentation` property from it, which you need to read from the `@param` tag. Of course as a new feature it will use more memory.

src/SignatureHelpProvider.php
vakata: Yes, but unfortunately most of the time when dealing with an unclosed brace the nearest node was very far from the actual position. As for the buffer - I will rewrite and use string operations, but I doubt I will have the time to do it tonight.

src/Definition.php
vakata: Yes, I currently populate only the label (using this string). I will add the documentation and construct `ParameterInformation` objects - of course that would mean an object for every parameter of every method / function in the project. I will test and let you know how it impacts memory usage.

src/DefinitionResolver.php
vakata: It is actually needed in order to provide signatureHelp when invoking a method on `$this`, and this is not part of the $this autocomplete PR, as it was not needed for autocomplete - different parts of the code are involved. I could move it to a separate PR?

src/SignatureHelpProvider.php
felixfbecker: Could you give an example?

vakata: I do not have the time to prepare an example at the moment. The problem was that the parser errors before reaching the cursor position, which means there is no node at that position - it should be easy to reproduce.

felixfbecker: In that case lets open issues for better error recovery at these places. A lot of improvements were already made in that area.

vakata: I just checked - the `methodNotClosed` test requires at least a couple of symbols of backtracking before `getNodeAtPosition` returns a node.

src/DefinitionResolver.php
```diff
@@ -134,7 +135,13 @@ public function createDefinitionFromNode(Node $node, string $fqn = null): Defini
         $def->parameters = [];
         if ($node instanceof Node\FunctionLike) {
             foreach ($node->getParams() as $param) {
-                $def->parameters[] = $this->prettyPrinter->prettyPrint([$param]);
+                if (!$param->getAttribute('parentNode')) {
```
vakata: @felixfbecker I could not figure out why those instances do not have a `parentNode` attribute - this is a temporary patch. Could you have a look?

src/Definition.php
vakata: @felixfbecker the impact was not significant

src/SignatureHelpProvider.php
vakata: Maybe this should remain open until I rewrite it to use `getContent` instead of a buffer?

src/Definition.php
felixfbecker: FYI we already keep `Hover` and `SymbolInformation` in memory. I thought about changing `Definition` so that it contains just the data necessary to construct those and then build them from this, but I would prefer to first implement the features and then improve perf.

If memory usage is too high the implementation would not save anything in `Definition` and load the document on demand.

felixfbecker: You can leave out the docs first but lets use objects from the start on.

src/DefinitionResolver.php
felixfbecker: Okay. If you do it in a separate PR I think I would merge it straight away, change looks good.

src/Protocol/SignatureHelp.php
```diff
@@ -29,4 +29,16 @@ class SignatureHelp
      * @var int|null
      */
     public $activeParameter;
+
+    /**
+     * @param SignatureInformation[] $signatures      The signatures.
+     * @param bool                   $activeSignature The active signature.
+     * @param bool                   $activeParameter The active parameter of the active signature.
+     */
+    public function __construct(array $signatures = [], int $activeSignature = 0, int $activeParameter = 0)
```
felixfbecker: Use `null` as default value for the last two params because they are optional on the object.

src/Protocol/SignatureHelp.php
```diff
@@ -29,4 +29,16 @@ class SignatureHelp
      * @var int|null
      */
     public $activeParameter;
+
+    /**
+     * @param SignatureInformation[] $signatures      The signatures.
+     * @param bool                   $activeSignature The active signature.
+     * @param bool                   $activeParameter The active parameter of the active signature.
```
felixfbecker: `int|null`

src/Protocol/SignatureInformation.php
```diff
@@ -31,4 +31,16 @@ class SignatureInformation
      * @var ParameterInformation[]|null
      */
     public $parameters;
+
+    /**
+     * @param string                 $label         The label of this signature. Will be shown in the UI.
+     * @param string|null            $documentation The human-readable doc-comment of this signature.
+     * @param ParameterInformation[] $parameters    The parameters of this signature.
+     */
+    public function __construct(string $label = null, string $documentation = null, array $parameters = [])
```
felixfbecker: Use `null` as default for `$parameters`

src/Protocol/SignatureInformation.php
```diff
@@ -31,4 +31,16 @@ class SignatureInformation
      * @var ParameterInformation[]|null
      */
     public $parameters;
+
+    /**
+     * @param string                 $label         The label of this signature. Will be shown in the UI.
+     * @param string|null            $documentation The human-readable doc-comment of this signature.
+     * @param ParameterInformation[] $parameters    The parameters of this signature.
```
felixfbecker: adapt type for `$parameters`

src/Definition.php
vakata: It is already done, and the docs are in place as well. I added a comment for you though - `Param` nodes do not have a `parentNode` attribute - I could not fix it - the attribute was added in `ReferenceAdder` but then was not available at a later point.

src/SignatureHelpProvider.php
felixfbecker: It is weird to use the parser for completion, but the lexer/content for signature help.
If error recovery works for completion for methods, it should work for the parameters too, right?
What is the emitted AST at the point where the parser "fails" for the signature help use case?

vakata: The node at the requested position is `null`, after backtracking 31 symbols (each of those positions is still `null`) we get to an `Error` node, whose parent is a `PropertyFetch` node. The `$var` property of the fetch is populated, but the `$name` property is not (it is the `Error` node). This is why I used string operations to get the method being called, and then - the `Lexer` with some custom logic on top to get the parameter count. 
As I said in the Signature Help issue - it is a messy solution, but it is the only one I could come up with. I am sure someone with superior skills can take it a few steps further to something elegant and more robust :)

src/DefinitionResolver.php
felixfbecker: Every node except the root node should have a parentNode
Otherwise there must be a bug in NodeVisitor or ReferencesAdder

src/Definition.php
vakata: @felixfbecker anything I can help with here?

src/Protocol/SignatureInformation.php
```diff
@@ -33,9 +33,9 @@ class SignatureInformation
...
 
     /**
-     * @param string                 $label         The label of this signature. Will be shown in the UI.
-     * @param string|null            $documentation The human-readable doc-comment of this signature.
-     * @param ParameterInformation[] $parameters    The parameters of this signature.
+     * @param string                      $label         The label of this signature. Will be shown in the UI.
+     * @param string|null                 $documentation The human-readable doc-comment of this signature.
+     * @param ParameterInformation[]|null $parameters    The parameters of this signature.
      */
     public function __construct(string $label = null, string $documentation = null, array $parameters = [])
```
felixfbecker: default value for `$parameters` must be `null`

vakata:
I am closing this as this solution will now longer work as it depends on the Lexer from nikic's PHP parser.

PR #255, opened 2017-01-24
Make Index an EventEmitter

### LINKED ISSUE #239: Make indexing more evented
Currently `initialize` blocks sending a response until indexing finished.
It should return instantly and kick off indexing in the background.
Requests that return a single result like hover or j2d should return a result if it is available, otherwise subscribe to new additions to the index and retry. It should also subscribe to indexing being finished, and then return an empty result.
Other methods that return multiple results like references or symbols should defer their response until indexing finished. In the future they might stream their result.

felixfbecker:
Closes #239 
This changes the whole indexing process to be async again. The `Index` is now an `EventEmitter` that emits two events:
- `definition-added` when a Definition was added
- `complete` when the indexing has completed (triggered through `setComplete()`)

`isCompete()` returns the current status.

All methods that return multiple results that depend on the project index (`workspace/symbols`, `textDocument/references`, `workspace/xreferences`) now defer the response until indexing finished.

All methods that need to resolve a single definition (`textDocument/(x)definition`, `textDocument/hover`) check the index for a result, then if no result was found and indexing did not finish yet, check again on every added definition.

Since `LanguageServer::index` is now executed async, adds new `beforeIndex` hook.

### DIFF
--- a/src/Index/Index.php
+++ b/src/Index/Index.php
@@ -45,6 +45,9 @@ class Index implements ReadableIndex
      */
     public function setComplete()
     {
+        if (!$this->isStaticComplete()) {
+            $this->setStaticComplete();
+        }
         $this->complete = true;
         $this->emit('complete');
     }
--- a/src/Index/Index.php
+++ b/src/Index/Index.php
@@ -34,6 +34,11 @@ class Index implements ReadableIndex
     private $complete = false;
 
     /**
+     * @var bool
+     */
+    private $staticComplete = false;
+
+    /**
      * Marks this index as complete
      *
      * @return void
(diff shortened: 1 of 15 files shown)


src/Index/Index.php
```diff
@@ -26,6 +29,58 @@ class Index implements ReadableIndex
...
+    }
+
+    /**
+     * Marks this index as complete for static definitions and references
+     *
+     * @return void
+     */
+    public function setStaticComplete()
+    {
+        $this->complete = true;
```
eseliger: `staticComplete`

src/Server/Workspace.php
```diff
@@ -61,17 +62,23 @@ public function __construct(ProjectIndex $index, DependenciesIndex $dependencies
...
-    public function symbol(string $query): array
+    public function symbol(string $query): Promise
     {
-        $symbols = [];
-        foreach ($this->index->getDefinitions() as $fqn => $definition) {
-            if ($query === '' || stripos($fqn, $query) !== false) {
-                $symbols[] = $definition->symbolInformation;
+        return coroutine(function () use ($query) {
+            // Wait until indexing for definitions finished
+            if (!$this->index->isComplete()) {
```
eseliger: `isStaticComplete`?

PR #260, opened 2017-01-26
Caching

### LINKED ISSUE #47: Cache indexing result

felixfbecker:
Closes #47 

Implements the [cache extension](https://github.com/sourcegraph/language-server-protocol/blob/68c291281db65eb41ad9890e71c962bda92b5e03/extension-chache.md) and alternatively falls back to the file system.
Cache directory is `~/.phpls` on Linux/Mac and `%LOCALAPPDATA%/PHP Language Server` on Windows.

The index for each dependency gets serialized and saved in the cache, keyed by package name and version.

Indexing is now done by the `Indexer` class.

### DIFF
--- a/src/Cache/ClientCache.php
+++ b/src/Cache/ClientCache.php
@@ -27,7 +27,9 @@ class ClientCache implements Cache
      */
     public function get(string $key): Promise
     {
-        return $this->client->xcache->get($key)->then('unserialize')->otherwise(function () {});
+        return $this->client->xcache->get($key)->then('unserialize')->otherwise(function () {
+            // Ignore
+        });
     }
 
     /**
@@ -39,6 +41,8 @@ class ClientCache implements Cache
      */
     public function set(string $key, $value): Promise
     {
-        return $this->client->xcache->set($key, serialize($value))->otherwise(function () {});
+        return $this->client->xcache->set($key, serialize($value))->otherwise(function () {
+            // Ignore
+        });
     }
 }
--- a/src/Cache/ClientCache.php
+++ b/src/Cache/ClientCache.php
@@ -1,0 +1,44 @@
+<?php
+declare(strict_types = 1);
+
+namespace LanguageServer\Cache;
+
+use LanguageServer\LanguageClient;
+use Sabre\Event\Promise;
+
+/**
+ * Caches content through a xcache/* requests
+ */
+class ClientCache implements Cache
+{
+    /**
+     * @param LanguageClient $client
+     */
+    public function __construct(LanguageClient $client)
+    {
+        $this->client = $client;
+    }
+
+    /**
+     * Gets a value from the cache
+     *
+     * @param string $key
+     * @return Promise <mixed>
+     */
+    public function get(string $key): Promise
+    {
+        return $this->client->xcache->get($key)->then('unserialize')->otherwise(function () {});
+    }
+
+    /**
+     * Sets a value in the cache
+     *
+     * @param string $key
+     * @param mixed  $value
+     * @return Promise
+     */
+    public function set(string $key, $value): Promise
+    {
+        return $this->client->xcache->set($key, serialize($value))->otherwise(function () {});
+    }
+}
--- a/src/Cache/FileSystemCache.php
+++ b/src/Cache/FileSystemCache.php
@@ -18,10 +18,12 @@ class FileSystemCache implements Cache
 
     public function __construct()
     {
-        if (PHP_OS === 'Windows') {
-            $this->cacheDir = getenv('LOCALAPPDATA') . '\\PHP Language Server\\';
+        if (PHP_OS === 'WINNT') {
+            $this->cacheDir = $_ENV['LOCALAPPDATA'] . '\\PHP Language Server\\';
+        } else if (!empty($_ENV['XDG_CACHE_HOME'])) {
+            $this->cacheDir = $_ENV['XDG_CACHE_HOME'] . '/phpls/';
         } else {
-            $this->cacheDir = getenv('HOME') . '/.phpls/';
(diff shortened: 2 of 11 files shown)


src/Cache/ClientCache.php
```diff
@@ -0,0 +1,48 @@
...
+    /**
+     * Sets a value in the cache
+     *
+     * @param string $key
+     * @param mixed  $value
+     * @return Promise
+     */
+    public function set(string $key, $value): Promise
+    {
+        return $this->client->xcache->set($key, serialize($value))->otherwise(function () {
```
staabm: Serialze has a lot of security implications and is slow. You should consider using e.g. json_encode, igbinary or similar

felixfbecker: What security implications does it have? `json_encode` would be magnitudes _slower_ because it does not store the class info, I would have to copy all properties into class instances. `serialize` output is also much smaller than `json_encode`.

staabm: serialize had dozens of security related problems in the past
https://www.google.de/search?q=php+serialize+cve

using serialize() will also mean your cache contents will be coupled to the class definitions (and future updates can make the contents inconsistent with future class definitions).

staabm: what do you mean by "the class info"?

felixfbecker:
```php
class MyClass {
  myMethod() {}
}
json_decode(json_encode(new MyClass))->myMethod() // not defined
```

felixfbecker: Re: Security. The security concerns all revolve around `unserialize` triggering autoloading and `__wakeup` calls of unwanted classes, if the serialized content has been tempered with and then relies on the presence of a class that would actually cause harm. For this, you would need to have access to the cache. If an attacker has access to the `~/.phpls`, he has write access to your user folder already, which means he could do a lot worse stuff than tampering with a PHP LS cache. And he also already has access to `~/.vscode/extensions`, which means instead of trying to inject an attack in the cache, he could just modify the PHP source in place, which would be much easier, can do a lot more harm and there is nothing we can do against it.

If, on the other hand, the cache is stored inside the client, then it is his responsibility to secure it.

> using serialize() will also mean your cache contents will be coupled to the class definitions (and future updates can make the contents inconsistent with future class definitions).

That's why all cache items are prefixed with a version. The cache format will eventually change at some point, whether I use JSON or `serialize`, it needs to be versioned.

staabm: you can either do stuff on your own:
https://3v4l.org/5Y60r

or use libs which provide this kind of functionality, e.g:
http://jmsyst.com/libs/serializer

Re Security: your class doesnt know whether the content beeing stored is provided by user-input or not.

> That's why all cache items are prefixed with a version. The cache format will eventually change at some point, whether I use JSON or serialize, it needs to be versioned.
nice!

felixfbecker: I am already using JSONMapper in the project for all LSP communications. _But it is slow_. Because you need to parse JSON, then iterate over all the properties _recursively_, use reflection to parse and check the `@var` tags of the classes to see what type they are, instantiate that class (which, btw, is also remote code execution), copy all properties values over, continue. `unserialize` is much, much faster for this.

> Re Security: your class doesnt know whether the content beeing stored is provided by user-input or not.

If you are referring to the `Cache` class, no, and it's not its job. Its job is to save and get cache items in serialized form. It's not its job to implement any security layer. The cache store is trusted, it is a database, _not_ user input. You don't sanitize values that come from your own database, you secure your database properly.

staabm: I guess we can agree to disagree ;-).

felixfbecker: The mapping with the algorithm I described is what makes it slow. And that is needed, because I need class instances of `Definition`, `Index` etc, not `\stdClass`

jrobeson:
can the cache directory be configurable? I'd like to use my linux native `$XDG_CACHE_HOME` (usually `~/.cache`.

felixfbecker:
@jrobeson There isn't any configuration in the language server, but if the env var is defined, I would default it to `$XDG_CACHE_HOME/phpls`. Sounds good?

jrobeson:
I'd appreciate it, thanks :)

src/Indexer.php
```diff
@@ -0,0 +1,222 @@
...
+            $count = count($uris);
+            $startTime = microtime(true);
+            $this->client->window->logMessage(MessageType::INFO, "$count files total");
+
+            /** @var string[] */
+            $source = [];
+            /** @var string[][] */
+            $deps = [];
+            foreach ($uris as $uri) {
+                if ($this->composerLock !== null && preg_match('/\/vendor\/([^\/]+\/[^\/]+)\//', $uri, $matches)) {
```
wulftone: I think this directory should be pulled from `composer.json`--not everyone uses the default `./vendor/` directory and they are missing out on these wonderful cache benefits.

felixfbecker: Feel free to do a PR

src/Cache/FileSystemCache.php
```diff
@@ -0,0 +1,67 @@
...
+class FileSystemCache implements Cache
+{
+    /**
+     * @var string
+     */
+    public $cacheDir;
+
+    public function __construct()
+    {
+        if (PHP_OS === 'Windows') {
```
jens1o: Don't use PHP_OS here, since it is just the system php was built from, not the system it is running at the moment. Use PHP_EOL instead(`\r\n` is typical for Windows)

PR #269, opened 2017-02-02
Automatically index stubs on composer install

cweagans:
This will only run automatically when a developer runs composer install - this is not run for anyone that gets this package as a requirement of some other package (so projects that require the language server will still have to manually index the stubs).


cweagans:
I have no idea. It works locally. Might be an issue with Travis' composer version or something?

PR #272, opened 2017-02-03
nvim-cm-php-language-server is using php language server

roxma:


felixfbecker:
Is this a vim plugin? If yes, please write "Vim (extension name)"

roxma:
This is a neovim plugin. `Neovim (nvim-cm-php-language-server)`

PR #281, opened 2017-02-06
Read vendor directory from project's composer.json, if set.

wulftone:
If not set, default to "vendor".

This PR also includes a bugfix for finding the appropriate composer.json and composer.lock files.

This PR does not address an existing issue with files that are included in composer.json via the "repositories" mechanism with a type of "package" mechanism, e.g.:

```json
        {
            "type": "package",
            "package": {
                "name": "ezyang/htmlpurifier",
                "version": "4.6.0",
                "dist": {
                    "url": "https://github.com/ezyang/HTMLPurifier/archive/v4.6.0.zip",
                    "type": "zip"
                }
            }
        }
```

These are not currently cached.  Those files are stored in the `[vendor]/composer/[some hash]/[packageName]-[packageVersion]` directory.  A future enhancement could cache these files as well.

### DIFF
--- a/src/Index/ProjectIndex.php
+++ b/src/Index/ProjectIndex.php
@@ -22,10 +22,11 @@ class ProjectIndex extends AbstractAggregateIndex
      */
     private $sourceIndex;
 
-    public function __construct(Index $sourceIndex, DependenciesIndex $dependenciesIndex)
+    public function __construct(Index $sourceIndex, DependenciesIndex $dependenciesIndex, \stdClass $composerJson = null)
     {
         $this->sourceIndex = $sourceIndex;
         $this->dependenciesIndex = $dependenciesIndex;
+        $this->composerJson = $composerJson;
         parent::__construct();
     }
 
@@ -43,7 +44,7 @@ class ProjectIndex extends AbstractAggregateIndex
      */
     public function getIndexForUri(string $uri): Index
     {
-        if (preg_match('/\/vendor\/([^\/]+\/[^\/]+)\//', $uri, $matches)) {
+        if (\LanguageServer\uriInVendorDir($this->composerJson, $uri, $matches)) {
             $packageName = $matches[1];
             return $this->dependenciesIndex->getDependencyIndex($packageName);
         }
--- a/src/Index/ProjectIndex.php
+++ b/src/Index/ProjectIndex.php
@@ -3,6 +3,8 @@ declare(strict_types = 1);
 
 namespace LanguageServer\Index;
 
+use function LanguageServer\getPackageName;
+
 /**
  * A project index manages the source and dependency indexes
  */
@@ -44,8 +46,8 @@ class ProjectIndex extends AbstractAggregateIndex
      */
     public function getIndexForUri(string $uri): Index
     {
-        if (\LanguageServer\uriInVendorDir($this->composerJson, $uri, $matches)) {
-            $packageName = $matches[1];
+        $packageName = getPackageName($this->composerJson, $uri);
+        if ($packageName) {
             return $this->dependenciesIndex->getDependencyIndex($packageName);
         }
         return $this->sourceIndex;
--- a/src/Index/ProjectIndex.php
+++ b/src/Index/ProjectIndex.php
@@ -46,7 +46,7 @@ class ProjectIndex extends AbstractAggregateIndex
      */
     public function getIndexForUri(string $uri): Index
     {
-        $packageName = getPackageName($this->composerJson, $uri);
+        $packageName = getPackageName($uri, $this->composerJson);
         if ($packageName) {
             return $this->dependenciesIndex->getDependencyIndex($packageName);
         }
--- a/src/Indexer.php
+++ b/src/Indexer.php
@@ -186,7 +186,7 @@ class Indexer
                         $this->client->window->logMessage(MessageType::INFO, "Storing $packageKey in cache");
                         $this->cache->set($cacheKey, $index);
                     } else {
-                        $this->client->window->logMessage(MessageType::INFO, "Cannot cache $packageName, cache key was null. Either you are using a 'dev-' version or your composer.lock is missing references.");
+                        $this->client->window->logMessage(MessageType::WARNING, "Could not compute cache key for $packageName");
                     }
                 }
             }
--- a/src/Indexer.php
+++ b/src/Indexer.php
@@ -117,7 +117,10 @@ class Indexer
             /** @var string[][] */
             $deps = [];
 
-            $vendorDir = str_replace('/', '\/', @$this->composerJson->config->{'vendor-dir'} ?: 'vendor');
+            $vendorDir = isset($this->composerJson->config->{'vendor-dir'}) ?
+                $this->composerJson->config->{'vendor-dir'}
+                : 'vendor';
+            $vendorDir = str_replace('/', '\/', $vendorDir);
             $this->client->window->logMessage(MessageType::INFO, "Vendor dir: $vendorDir");
 
             foreach ($uris as $uri) {
--- a/src/Indexer.php
+++ b/src/Indexer.php
@@ -118,9 +118,9 @@ class Indexer
             $deps = [];
 
             foreach ($uris as $uri) {
-                if ($this->composerLock !== null && uriInVendorDir($this->composerJson, $uri, $matches)) {
+                $packageName = getPackageName($this->composerJson, $uri);
+                if ($this->composerLock !== null && $packageName) {
                     // Dependency file
-                    $packageName = $matches[1];
                     if (!isset($deps[$packageName])) {
                         $deps[$packageName] = [];
                     }
--- a/src/Indexer.php
+++ b/src/Indexer.php
@@ -60,6 +60,11 @@ class Indexer
     private $composerLock;
 
     /**
+     * @var \stdClasss
+     */
+    private $composerJson;
+
+    /**
      * @param FilesFinder       $filesFinder
      * @param string            $rootPath
      * @param LanguageClient    $client
@@ -77,7 +82,8 @@ class Indexer
         DependenciesIndex $dependenciesIndex,
         Index $sourceIndex,
         PhpDocumentLoader $documentLoader,
-        \stdClass $composerLock = null
+        \stdClass $composerLock = null,
+        \stdClass $composerJson = null
     ) {
         $this->filesFinder = $filesFinder;
         $this->rootPath = $rootPath;
@@ -87,6 +93,7 @@ class Indexer
         $this->sourceIndex = $sourceIndex;
         $this->documentLoader = $documentLoader;
         $this->composerLock = $composerLock;
+        $this->composerJson = $composerJson;
     }
 
     /**
@@ -109,8 +116,12 @@ class Indexer
             $source = [];
             /** @var string[][] */
             $deps = [];
+
+            $vendorDir = str_replace('/', '\/', @$this->composerJson->config->{'vendor-dir'} ?: 'vendor');
+            $this->client->window->logMessage(MessageType::INFO, "Vendor dir: $vendorDir");
+
             foreach ($uris as $uri) {
-                if ($this->composerLock !== null && preg_match('/\/vendor\/([^\/]+\/[^\/]+)\//', $uri, $matches)) {
+                if ($this->composerLock !== null && preg_match("/\/$vendorDir\/([^\/]+\/[^\/]+)\//", $uri, $matches)) {
                     // Dependency file
                     $packageName = $matches[1];
                     if (!isset($deps[$packageName])) {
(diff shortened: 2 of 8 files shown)


src/Indexer.php
```diff
@@ -109,8 +116,12 @@ public function index(): Promise
             $source = [];
             /** @var string[][] */
             $deps = [];
+
+            $vendorDir = str_replace('/', '\/', @$this->composerJson->config->{'vendor-dir'} ?: 'vendor');
```
felixfbecker: don't use the `@` operator, check this with `if`/`isset`

src/Indexer.php
```diff
@@ -174,6 +185,8 @@ public function index(): Promise
                     if ($cacheKey !== null) {
                         $this->client->window->logMessage(MessageType::INFO, "Storing $packageKey in cache");
                         $this->cache->set($cacheKey, $index);
+                    } else {
+                        $this->client->window->logMessage(MessageType::INFO, "Cannot cache $packageName, cache key was null. Either you are using a 'dev-' version or your composer.lock is missing references.");
```
felixfbecker: Use a warning and shorten text to `Could not compute cache key for $packageName`

src/LanguageServer.php
```diff
@@ -206,6 +206,13 @@ public function initialize(ClientCapabilities $capabilities, string $rootPath =
                 // Find composer.json
                 if ($this->composerJson === null) {
                     $composerJsonFiles = yield $this->filesFinder->find(Path::makeAbsolute('**/composer.json', $rootPath));
+
+                    // If we sort our findings by string length (shortest to longest),
+                    // the first entry will be the project's root composer.json.
```
felixfbecker: This is not true
```
/myproject/http_rest_api_backend/composer.json
/myproject/vendor/psr/log/composer.json
```

wulftone: Good call, should have thought about it for more than two seconds.  I think `substr_count($a, '/')` will do it correctly?

src/Indexer.php
wulftone: I thought including a potential reason would help some people.  I can shorten it though.

felixfbecker: Logs are meant for developers and admins

src/LanguageServer.php
felixfbecker: Yes, that would work! 👍 
But you need to apply it to the `path` component of the URI, otherwise a URL with host will have 2 more slashes than a URN. Use `Uri\parse()`

wulftone: I don't see a `Uri` class included.  Do you have a parsing library to recommend adding to the project?

felixfbecker: You will need to add `use Sabre\Uri`

src/utils.php
```diff
@@ -7,6 +7,7 @@
 use InvalidArgumentException;
 use PhpParser\Node;
 use Sabre\Event\{Loop, Promise, EmitterInterface};
+use function Sabre\Uri\parse;
```
felixfbecker: please only alias `Sabre\Uri` and reference with `Uri\parse()`, otherwise it is harder to understand what is going on.

src/utils.php
```diff
@@ -131,3 +132,19 @@ function stripStringOverlap(string $a, string $b): string
...
+ * in ascending order.
+ *
+ * Example:
+ *     usort($uriList, 'LanguageServer\sortByLeastSlashes');
+ *
+ * @param $a string
+ * @param $b string
+ * @return integer
+ */
+function sortByLeastSlashes($a, $b)
```
felixfbecker: make this function take the array and call `usort` on it
rename to `sortUrisLevelOrder`, it makes it more clear what it does

src/LanguageServer.php
```diff
@@ -25,6 +25,7 @@
 use Throwable;
 use Webmozart\PathUtil\Path;
 use Sabre\Uri;
+use function LanguageServer\sortByLeastSlashes;
```
felixfbecker: the function is in the same namespace

src/utils.php
```diff
@@ -137,14 +137,12 @@ function stripStringOverlap(string $a, string $b): string
...
  * in ascending order.
  *
- * Example:
- *     usort($uriList, 'LanguageServer\sortByLeastSlashes');
- *
- * @param $a string
- * @param $b string
- * @return integer
+ * @param array
+ * @return array
```
felixfbecker: `@return void`

src/utils.php
```diff
@@ -137,14 +137,12 @@ function stripStringOverlap(string $a, string $b): string
...
  */
-function sortByLeastSlashes($a, $b)
+function sortUrisLevelOrder(&$uriList)
 {
-    return substr_count(parse($a)['path'], '/') - substr_count(parse($b)['path'], '/');
+    usort($uriList, function ($a, $b) {
+        return substr_count(Uri\parse($a)['path'], '/') - substr_count(Uri\parse($b)['path'], '/');
```
felixfbecker: only parse the URI once

wulftone:
There are a few places where `vendor` occurs, but I'm at a loss for a good strategy to deal with all of them all.  Some are embedded in interesting places and getting a reference to `composerJson` in there is not straightforward.  I'd like to centralize the logic if possible, but if you could give me a bit of architectural direction that would be quite helpful...

wulftone:
```

src/PhpDocument.php
231:        return strpos($path, '/vendor/') !== false;

src/Server/TextDocument.php
412:            if (preg_match('/\/vendor\/([^\/]+\/[^\/]+)\//', $def->symbolInformation->location->uri, $matches) && $this->composerLock !== null) {

src/Server/Workspace.php
125:                    preg_match('/\/vendor\/([^\/]+\/[^\/]+)\//', $def->symbolInformation->location->uri, $matches);
```
I think TextDocument will be straightforward but not PhpDocument.  Workspace might be doable simply, too.  Looking into it now.

felixfbecker:
The `isVendored()` method can be removed in favor of just checking from the outside on `getUri()`. The other places shouldn't be too hard to inject an instance of the composer.json file

src/utils.php
```diff
@@ -138,11 +138,22 @@ function stripStringOverlap(string $a, string $b): string
...
+ * @return void
  */
 function sortUrisLevelOrder(&$uriList)
 {
-    usort($uriList, function ($a, $b) {
-        return substr_count(Uri\parse($a)['path'], '/') - substr_count(Uri\parse($b)['path'], '/');
+    // Parse URIs so we are not continually parsing them while sorting
+    $parsedUriList = array_map(function ($uri) {
+        return Uri\parse($uri);
+    }, $uriList);
```
felixfbecker: I'm sorry, I had a brain fart in my last review. I though `$a` and `$b` were the same, but they are not.
This is just micro optimisation \

wulftone: Haha ok I was wondering like, "does he have a use-case with a million files?" 🎈

src/Index/ProjectIndex.php
```diff
@@ -43,7 +44,7 @@ protected function getIndexes(): array
      */
     public function getIndexForUri(string $uri): Index
     {
-        if (preg_match('/\/vendor\/([^\/]+\/[^\/]+)\//', $uri, $matches)) {
+        if (\LanguageServer\uriInVendorDir($this->composerJson, $uri, $matches)) {
```
felixfbecker: alias the function

src/utils.php
```diff
@@ -146,3 +146,47 @@ function sortUrisLevelOrder(&$uriList)
...
+/**
+ * Check a given URI against the composer.json to see if it
+ * is a vendored URI
+ *
+ * @param \stdClass|null $composerJson
+ * @param string         $uri
+ * @param array          $matches
+ * @return int
+ */
+function uriInVendorDir(\stdClass $composerJson = null, string $uri, &$matches): int
```
felixfbecker: can you rename this to `getPackageName` and make it return a string or `null`?

tests/PhpDocumentTest.php
```diff
@@ -42,18 +42,18 @@ public function testGetNodeAtPosition()
     public function testIsVendored()
     {
         $document = $this->createDocument('file:///dir/vendor/x.php', "<?php\n$\$a = new SomeClass;");
-        $this->assertEquals(true, $document->isVendored());
+        $this->assertEquals(true, \LanguageServer\isVendored($document));
```
felixfbecker: alias the function

src/utils.php
```diff
@@ -146,3 +146,47 @@ function sortUrisLevelOrder(&$uriList)
...
+/**
+ * Helper function to get the vendor directory from composer.json
+ * or default to 'vendor'
+ *
+ * @param \stdClass|null $composerJson
+ * @return string
+ */
+function getVendorDir(\stdClass $composerJson = null): string
+{
+    return isset($composerJson->config->{'vendor-dir'}) ?
```
felixfbecker: I think `config` could be undefined too.

wulftone: It should work correctly, see the following example:
```php
$ psysh
Psy Shell v0.7.1 (PHP 7.1.1 — cli) by Justin Hileman
>>> $a = new \stdClass;
=> {#197}
>>> $a->foo->bar = 1
PHP warning:  Creating default object from empty value on line 1
>>> $a
=> {#197
     +"foo": {#200
       +"bar": 1,
     },
   }
>>> isset($a->foo->bar);
=> true
>>> isset($a->foo->baz);
=> false
>>> isset($a->foz->baz);
=> false 
// isset operates just like @ and suppresses errors for the whole line (in parens anyway)
```

src/Server/Workspace.php
```diff
@@ -9,6 +9,7 @@
 use Sabre\Event\Promise;
 use function Sabre\Event\coroutine;
 use function LanguageServer\waitForEvent;
+use function LanguageServer\getPackageName;
```
felixfbecker: use a group use statement

src/utils.php
```diff
@@ -169,12 +169,13 @@ function isVendored(PhpDocument $document, \stdClass $composerJson = null): bool
  * @param \stdClass|null $composerJson
  * @param string         $uri
  * @param array          $matches
- * @return int
+ * @return string
  */
-function uriInVendorDir(\stdClass $composerJson = null, string $uri, &$matches): int
+function getPackageName(\stdClass $composerJson = null, string $uri): ?string
```
felixfbecker: I am pretty sure the return syntax you are using is only available in PHP7.1. That would be a breaking change, please omit the return type annotation

src/utils.php
```diff
@@ -169,12 +169,13 @@ function isVendored(PhpDocument $document, \stdClass $composerJson = null): bool
  * @param \stdClass|null $composerJson
  * @param string         $uri
  * @param array          $matches
- * @return int
+ * @return string
```
felixfbecker: this should be `@return string|null`

src/utils.php
```diff
@@ -169,12 +169,13 @@ function isVendored(PhpDocument $document, \stdClass $composerJson = null): bool
...
  */
-function uriInVendorDir(\stdClass $composerJson = null, string $uri, &$matches): int
+function getPackageName(\stdClass $composerJson = null, string $uri): ?string
 {
     $vendorDir = str_replace('/', '\/', getVendorDir($composerJson));
-    return preg_match("/\/$vendorDir\/([^\/]+\/[^\/]+)\//", $uri, $matches);
+    preg_match("/\/$vendorDir\/([^\/]+\/[^\/]+)\//", $uri, $matches);
+    return isset($matches[1]) ? $matches[1] : null;
```
felixfbecker:
```php
return $matches[1] ?? null;
```

src/Server/TextDocument.php
```diff
@@ -134,7 +134,7 @@ public function documentSymbol(TextDocumentIdentifier $textDocument): Promise
     public function didOpen(TextDocumentItem $textDocument)
     {
         $document = $this->documentLoader->open($textDocument->uri, $textDocument->text);
-        if (!$document->isVendored()) {
+        if (!\LanguageServer\isVendored($document, $this->composerJson)) {
```
felixfbecker: alias the function

src/utils.php
```diff
@@ -131,3 +132,62 @@ function stripStringOverlap(string $a, string $b): string
...
+/**
+ * Check a given URI against the composer.json to see if it
+ * is a vendored URI
+ *
+ * @param \stdClass|null $composerJson
+ * @param string         $uri
+ * @param array          $matches
+ * @return string
+ */
+function getPackageName(\stdClass $composerJson = null, string $uri): ?string
```
felixfbecker: can you change this to take `$composerJson` as second parameter, so it's consistent with `isVendored()`?

wulftone:
Lotta commits in here, might want to squash and merge so it's cleaner (I usually rebase and squash for my own work)

felixfbecker:
Thanks!

wulftone:
Thank you!  Happy to collaborate with a responsive maintainer.

PR #283, opened 2017-02-06
Use php_uname() instead of PHP_OS

felixfbecker:
`PHP_OS` only shows the OS PHP was built on.

@jens1o

### DIFF
--- a/src/Cache/FileSystemCache.php
+++ b/src/Cache/FileSystemCache.php
@@ -18,7 +18,7 @@ class FileSystemCache implements Cache
 
     public function __construct()
     {
-        if (PHP_OS === 'WINNT') {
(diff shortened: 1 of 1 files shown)


felixfbecker:
@jens1o This has nothing to do with the line ending. You could have an OS that uses CRLF as a line ending and is not Windows.

jens1o:
Or some operating system that might fake it's own name 😉

felixfbecker:
That's like being asked about your allergies, lying and then complaining when you get an allergic shock from the food you are served.

staabm:
a well-used hack is checking the DIRECTORY_SEPARATOR not PHP_EOL.

felixfbecker:
Why does everyone want me to use hacks instead of functions designed for this 😄

staabm:
usually for perf reasons.

Issue #159, opened 2016-11-18
Excluding files from indexing


felixfbecker:
For example, node_modules

Alanaktion:
I would particularly like the ability to exclude certain directories from the Problems list (e.g. `vendor`), while retaining Go to Definition functionality. I'm not sure if that would be considered a separate issue.

felixfbecker:
#164 
Warnings from vendor are already ignored.

ryanbowden:
Really need this feature hope it comes soon. to make the searching much quicker.

felixfbecker:
@ryanbowden feel free to do a PR

damienflament:
The main use case is to ignore the cache files.

damienflament:
After seeing the code, the Indexer have to get the excluded path from options and give it to the FileFinder as second parameter. The FileFinder exclude files matching the given Globs.

I think it's the simplest way to do it as trying to compile a glob pattern using the main pattern (which will be dynamically generated due to another feature request) and the excluding patterns seems complicated.

I need to make a the development environment and try to submit a PR! When I got free time...

alex-pex:
from #381:
I'd like php intellisense to read files.exclude setting + php.excludeFiles (or a similar new config entry) with this value by default
```json
{
  "php.excludeFiles": {
    "**/node_modules": true"
  }
}
```

jens1o:
As soon as the large pr has been merged, I planned to do this, but there are some things that needs to be implemented first. (For example the client needs to send the settings to the server.)

dominikzogg:
https://github.com/atom/ide-php/issues/20

AeonDigital:
For those who do not want to wait for an update and do not mind making a "in the hand" follows a simple and homely solution:


1. Open "C:\Users\[user]\.vscode\extensions\felixfbecker.php-intellisense-1.5.1" in vscode.  
2. Open "\out\extension.js" and edit:
    look for line 22 and add the new instruction below:
    > 22: const memoryLimit = conf.get('memoryLimit') || '-1';
    > __23: const intellisense = conf.get('intellisense') || {'excludePaths': ''};    // new instruction__
    > 24: if (memoryLimit !== '-1' && !/^\d+[KMG]?$/.exec(memoryLimit)) { ...

    So, go to line 68 and add this:
    > 68: args.push('--memory-limit=' + memoryLimit);
    > __69: args.push('--exclude-paths=' + intellisense.excludePaths);    // new instruction__
[...]

PR #308, opened 2017-02-18
allow configurable file extension for indexing

JSteitz:
Requires the PR from vscode-php-intellisense felixfbecker/vscode-php-intellisense#77


JSteitz:
If everything is fine I will do the other 3 requests :)

src/Indexer.php
```diff
@@ -104,8 +112,8 @@ public function __construct(
     public function index(): Promise
     {
         return coroutine(function () {
-
-            $pattern = Path::makeAbsolute('**/*.php', $this->rootPath);
+            $fileTypes = implode(',', $this->options->fileTypes);
+            $pattern = Path::makeAbsolute('**/*{' . $fileTypes . '}', $this->rootPath);
```
jens1o: maybe check first if the extensions are still valid?

src/Indexer.php
```diff
@@ -65,14 +65,20 @@ class Indexer
...
+    /**
+     * @param FilesFinder            $filesFinder
+     * @param string                 $rootPath
+     * @param LanguageClient         $client
+     * @param Cache                  $cache
+     * @param DependenciesIndex      $dependenciesIndex
+     * @param Index                  $sourceIndex
+     * @param PhpDocumentLoader      $documentLoader
+     * @param \stdClass|null         $composerLock
+     * @param IndexerOptions|null    $options
```
staabm: Wrong type

src/Options.php
```diff
@@ -9,29 +9,67 @@ class Options
...
     /**
-     * @param \stdClass|null $options
+     * @param \Traversable|\stdClass|array|null $options
      */
-    public function __construct(\stdClass $options = null)
+    public function __construct($options = null)
     {
         // Do nothing when the $options parameter is not an object
-        if (!is_object($options)) {
+        if (!is_object($options) && !is_array($options) && (!$options instanceof \Traversable)) {
```
jens1o: that won't work, you need to do it like this: `!($options instanceof \Traversable)`

src/Options.php
```diff
@@ -9,29 +9,67 @@ class Options
...
             return;
         }
 
-        $this->fileTypes = $options->fileTypes ?? $this->normalizeFileTypes($this->fileTypes);
+        foreach ($options as $option => $value) {
+            $method = 'set' . ucfirst($option);
+
+            call_user_func([$this, $method], $value);
```
jens1o: Check first wether the method exists?

JSteitz: thx will do it later when I'm back home again

src/Options.php
JSteitz: I'll will add tests for the Option class when I'm back

src/Options.php
```diff
@@ -9,29 +9,67 @@ class Options
...
+    /**
+     * Validate and set options for file types
+     *
+     * @param array $fileTypes List of file types
+     */
+    public function setFileTypes(array $fileTypes)
+    {
+        $fileTypes = filter_var_array($fileTypes, FILTER_SANITIZE_STRING);
+        $fileTypes =  filter_var($fileTypes, FILTER_CALLBACK, ['options' => [$this, 'filterFileTypes']]);
+        $fileTypes = array_filter($fileTypes);
```
jens1o: You need to provide an callback for `array_filter`

src/Options.php
```diff
@@ -9,29 +9,67 @@ class Options
...
+
+    /**
+     * Validate and set options for file types
+     *
+     * @param array $fileTypes List of file types
+     */
+    public function setFileTypes(array $fileTypes)
+    {
+        $fileTypes = filter_var_array($fileTypes, FILTER_SANITIZE_STRING);
+        $fileTypes =  filter_var($fileTypes, FILTER_CALLBACK, ['options' => [$this, 'filterFileTypes']]);
```
jens1o: I think it may be possible that you'll pass false into this method.

src/Options.php
JSteitz: With no callback it will remove all that are false, but I can one

src/Options.php
JSteitz: Yeah, did a manual test before I used it, will be tested later

src/Options.php
jens1o: https://secure.php.net/manual/en/function.array-filter.php says, that the second parameter must be provided. But you could use `strlen`, that will filter any false, empty strings or null.

src/Options.php
felixfbecker: I would rather not use setter methods at all and let `Options` have public attributes.

src/Options.php
jens1o: ah, thanks. Didn't knew that.

src/LanguageServer.php
```diff
@@ -167,11 +167,12 @@ public function __construct(ProtocolReader $reader, ProtocolWriter $writer)
      * @param ClientCapabilities $capabilities The capabilities provided by the client (editor)
      * @param string|null $rootPath The rootPath of the workspace. Is null if no folder is open.
      * @param int|null $processId The process Id of the parent process that started the server. Is null if the process has not been started by another process. If the parent process is not alive then the server should exit (see exit notification) its process.
+     * @param mixed $initializationOptions The options send from client to initialize the server
      * @return Promise <InitializeResult>
      */
-    public function initialize(ClientCapabilities $capabilities, string $rootPath = null, int $processId = null): Promise
+    public function initialize(ClientCapabilities $capabilities, string $rootPath = null, int $processId = null, $initializationOptions = null): Promise
```
felixfbecker: I didn't know vscode sends the settings as `initializationOptions`, I thought `didChangeConfiguration` was the only way. Is there a reason you didn't type `$initializationOptions` as a class (`Options`)?

src/Options.php
```diff
@@ -0,0 +1,75 @@
...
+        // Do nothing when the $options parameter is not an object
+        if (!is_object($options) && !is_array($options) && (!$options instanceof \Traversable)) {
+            return;
+        }
+
+        foreach ($options as $option => $value) {
+            $method = 'set' . ucfirst($option);
+
+            call_user_func([$this, $method], $value);
+        }
```
felixfbecker: I would rather let `JsonMapper` do this (like it does for all protocol structures) instead of doing this here

src/Options.php
JSteitz: @felixfbecker How would you validate/filter the option? Calling them extra or using `__set()`?

src/Options.php
JSteitz: Weird for me it says it is optional :/

> If no callback is supplied, all entries of array equal to FALSE (see converting to boolean) will be removed.

src/LanguageServer.php
JSteitz: Microsoft has a list with all methods and params: https://github.com/Microsoft/language-server-protocol/blob/master/versions/protocol-2-x.md#initialize

You only have to pass them from vscode as a property of ClientOptions

According o the Docs it can have any type.

src/Options.php
felixfbecker: I wouldn't. In the LS code I just don't assign invalid values and JSONMapper will throw if the `@var` tags don't match the type

src/LanguageServer.php
felixfbecker: I can't force every client to do this. This needs to happen through `onDidChangeConfiguration`.

JSteitz: The actual options are taken from the configuration and are just passed with the initialize method. felixfbecker/vscode-php-intellisense@59e6f2cc2f7047bf3990d7463ab546c89eb330da

So you want to move the indexing to `onDidChangeConfiguration`? Otherwise you would index 2x, because this event is sent after the intialize (by vscode).

src/Options.php
JSteitz: Will it notify the client about the invalid value type?
What about arrays that could have any values as type? E.g. `$fileTypes = ['.php', false]`

Since the options are coming from the client we should validate/filter them

src/LanguageServer.php
felixfbecker: Tricky. A client might never send `onDidChangeConfiguration`. But it should definitely be retriggered if the config is sent through that.

src/Options.php
felixfbecker: Sure, the exception can be caught and propagated to the client.
What do you mean arrays that can have any type as value? `$fileTypes` would be of type `string[]`

src/LanguageServer.php
JSteitz: I would open a new issue to discuss this part because it affects the indexing for many option combinations. And maybe delay it until we have implemented some of them.

For now it is okay, when the user has to reload to reindex with the changed options.
I have already some ideas but I need to collect more details for how the indexing will be affected.

felixfbecker: I would want at least an implementation of `onDidChangeConfiguration` that, if this setting changed, wipes the index and reindexes, because I don't want to add an important feature like this to the language server if it only works in VS Code

felixfbecker: But feel free to discuss in an issue

haringsrob:
This is a blocker for me, anything I can do here to speed it up? :)

JSteitz:
Code for the `didConfigurationChange` is done, but the event is somehow not triggering for me.
Didn't have enough time this week to investigate.

JSteitz:
@felixfbecker I implemented the `workspace/didChangeConfiguration`, but I'm stuck with the `Options` class not updating for the indexer.

I don't like the idea to have the options to be global, what's  your opinion on that?

src/Server/Workspace.php
```diff
@@ -200,4 +214,44 @@ public function xdependencies(): array
...
+     * @param Options|null $settings
+     */
+    public function didChangeConfiguration(Options $settings = null)
+    {
+        if ($settings === null) {
+            return;
+        }
+
+        $changedOptions = $this->getChangedOptions($settings);
+        $this->options = $settings;
```
felixfbecker: This should just do
```php
foreach (get_object_vars($settings) as $prop => $val) {
  $this->options->$prop = $val;
}
```
then the options will get updated.

Do you have any specific concerns over sharing the Options instance?

src/Options.php
```diff
@@ -0,0 +1,64 @@
...
+
+        $this->fileTypes = !empty($fileTypes) ? $fileTypes : $this->fileTypes;
+    }
+
+    /**
+     * Get list with options that affect the indexer
+     *
+     * @return array
+     */
+    public function getIndexerOptions(): array
```
felixfbecker: I would like to keep `Options` a dumb data object (like other protocol structures).
Checking if reindexing is needed should be done solely in `didChangeConfiguration`.

src/Server/Workspace.php
```diff
@@ -200,4 +214,44 @@ public function xdependencies(): array
...
+
+        if (!empty(array_intersect($changedOptions, $this->options->getIndexerOptions()))) {
```
felixfbecker: This looks overly complicated. All it needs to check atm is the attribute `fileTypes`.

src/Server/Workspace.php
```diff
@@ -200,4 +214,44 @@ public function xdependencies(): array
...
+     * Get a list with all options that changed since last time
+     *
+     * @param Options $settings
+     * @return array List with changed options
+     */
+    private function getChangedOptions(Options $settings): array
+    {
+        // squash nested array for comparing changed options
+        $old = array_map('json_encode', get_object_vars($this->options));
+        $new = array_map('json_encode', get_object_vars($settings));
```
felixfbecker: JSON encoding is not a good way to do deep comparison. PHP can do this natively with `==`.

src/Index/AbstractAggregateIndex.php
```diff
@@ -42,6 +42,9 @@ protected function registerIndex(ReadableIndex $index)
         $index->on('definition-added', function () {
             $this->emit('definition-added');
         });
+        $index->on('wipe', function() {
```
felixfbecker: Who subscribes to this event? What is it needed for?
As it is implemented, it will be triggered many times for a single `wipe()` call on an AggregateIndex.

JSteitz: Currently only the tests are listening to this event.

felixfbecker: The runtime code shouldn't emit events just so tests can listen to this - if a method call is to be asserted, the object should be mocked. Atm I do this manually in some tests, a better way would be to use a mocking library (PHPUnit or prophecy)

JSteitz: Then they are leftovers from a previous attempt. Will remove them.

src/Server/Workspace.php
JSteitz: Yes right now only the `fileTypes` ist implemented.
Later we will have more that trigger a reindex and this part will prevent reindexing more than once at a time.

`getIndexerOptions()` can move somewhere else or be defined as variable inside`didChangeConfiguration`.

src/Server/Workspace.php
JSteitz: Mostly because of side-effects since the server is running until closed.

### Proposals:

1. `Options` trait with method `setOptions` for dependencies
2. `Options` will be handled by `didChangeConfiguration` and they (the values) will be passed as parameter to methods 
> e.g. `$indexer->index($fileTypes, $maxFileSize);`

I like option 1 more then 2 because the `Workspace` does not need to know how the dependencies use them. He only needs to inform them about the new settings.

Either way the `Workspace` needs all classes as dependencies that require `Options`.

src/Server/Workspace.php
```diff
@@ -216,16 +216,38 @@ public function xdependencies(): array
...
      */
-    public function didChangeConfiguration(Options $settings = null)
+    public function didChangeConfiguration($settings = null)
     {
         if ($settings === null) {
             return;
         }
 
+        // VSC sends the settings with the config section as main key
+        if ($settings instanceof \stdClass && $settings->phpIntelliSense) {
```
JSteitz: Not happy with this part.
The documentation is lacking in which format the settings are passed to the server and why they do it that way.
With all the different tests I did, they always send it like that:

```
{
    "ExtensionConfigSectionName": { // defined in the package.json in the vscode extensions
        ... the actual settings ...
    }
}

File: src/Server/Workspace.php [hunk 15]
```diff
@@ -200,4 +214,75 @@ public function xdependencies(): array
         }
         return $dependencyReferences;
     }
+
+    /**
+     * Fires when client changes settings in the client
```
felixfbecker: Change this to the protocol wording:

> A notification sent from the client to the server to signal the change of configuration settings.

https://github.com/Microsoft/language-server-protocol/blob/master/protocol.md#didchangeconfiguration-notification

src/Server/Workspace.php
```diff
@@ -200,4 +214,75 @@ public function xdependencies(): array
...
+     *
+     * The default paramter type is Options but it also accepts different types
+     * which will be transformed on demand.
+     *
+     * Currently only the vscode format is supported
```
felixfbecker: What does this mean? Where is the problem in saying the PHP LS reads settings under the `php` key?

src/LanguageServer.php
```diff
@@ -190,6 +191,7 @@ public function initialize(ClientCapabilities $capabilities, string $rootPath =
             $this->projectIndex = new ProjectIndex($sourceIndex, $dependenciesIndex, $this->composerJson);
             $stubsIndex = StubsIndex::read();
             $this->globalIndex = new GlobalIndex($stubsIndex, $this->projectIndex);
+            $initializationOptions = $initializationOptions ?? new Options;
```
felixfbecker: This way `$initializationOptions` is not an instance of `Options` if provided

src/Server/Workspace.php
```diff
@@ -200,4 +214,75 @@ public function xdependencies(): array
...
+     *
+     * @param mixed|null $settings
+     * @return bool
```
felixfbecker: what is the return value for? A notification does not return anything

tests/Server/Workspace/DidChangeConfigurationTest.php
```diff
@@ -0,0 +1,153 @@
...
+    }
+
+    /**
+     * Verify that the required methods for a reindex are called
+     */
+    public function testFileTypesOptionTriggersAReindex()
+    {
+        $sourceIndex = new Index;
+        $dependenciesIndex = new DependenciesIndex;
+        $projectIndex = $this->getMockBuilder('LanguageServer\Index\ProjectIndex')
```
JSteitz: Ohh yeah my bad, forgot that we have `::class`

src/Server/Workspace.php
```diff
@@ -216,28 +216,55 @@ public function xdependencies(): array
...
+
         if ($settings === null) {
-            return;
+            return false;
+        }
+
+        // VSC sends the settings with the config section as main key
+        if ($settings instanceof \stdClass && $settings->phpIntelliSense) {
+            $mapper = new \JsonMapper();
+            $settings = $mapper->map($settings->phpIntelliSense, new Options);
```
JSteitz: Sure we can.

We only have to lookout for possible conflicts in the future when we use `php` as config section.

Possible config naming:

- `php.indexer.*`
- `php.completion.*` (possible name for a feature request, but can be ignored now)
- `php.intellisense.*` (previous name)

felixfbecker: `php.fileExtensions` should be fine. I don't think we should concern user with the concept of an "indexer"

src/Server/Workspace.php
JSteitz: I know and I was hesitating to add the return value, but for some tests I need them.
When you have another idea how I can test them, tell me.

see for example `testNoOptionPassed` or `testNoChangedOptions`

src/LanguageServer.php
JSteitz: Don't get you here, is this comment a part from another comment?

felixfbecker: If `$initializedOptions` is not `null`, it will be `stdClass` because it's not typed. I also think you don't account for the nesting under `->php` here

JSteitz: For `$initializationOptions` you have more freedom how and which options will be send.

The `$initializationOptions` parameter ist typed in `initialize()` and is only passed as is to `coroutine()`, so it should be enough only to check for `null`.

src/Server/Workspace.php
```diff
@@ -216,28 +216,55 @@ public function xdependencies(): array
...
+        }
+
+        if (!($settings instanceof Options)) {
```
JSteitz: Do you have to tell him to that?
Because for me JsonMapper never threw an exception when the schema didn't match.
It silently ignored invalid properties and used the defaults in the `Options`.

src/LanguageServer.php
felixfbecker: You pass this variable to the `Indexer` further down, which accepts `Options`, not `stdClass`

src/Server/Workspace.php
felixfbecker: You shouldn't add return values just to do assertions when what you really want to test is that certain side effects happen. You should mock classes and then assert that a method is called or not.

src/LanguageServer.php
JSteitz: I'm really sorry but I don't see a problem here.
`$initializationOptions` is always of type `Options`.

At which point will it be a `stdClass`?

src/Server/Workspace.php
felixfbecker: What case exactly did not throw an exception? It could be because you typed `setFileTypes()` as taking `array`, not `string[]`

src/LanguageServer.php
felixfbecker: Oh sorry, I think I missed the type annotation because I looked at the closure `use`. But shouldn't this be the same object as passed to `didChangeConfiguration`? And will this break passing custom properties from the client in `initializationOptions`?

JSteitz: [`initializationOptions`](https://github.com/JSteitz/vscode-php-intellisense/blob/74be99d15703b16b810334b76e53491bc1bc4a4c/src/extension.ts#L101) can be completely different to [`didChangeConfiguration`](https://github.com/JSteitz/vscode-php-intellisense/blob/74be99d15703b16b810334b76e53491bc1bc4a4c/src/extension.ts#L97).
And actually they are not the same in this PR. While `initializationOptions` returns only defined options in the extension, `didChangeConfiguration` will return the whole section with the section name as main key.


**initializationOptions**
```
{
  "fileTypes": [
    ".php"
  ]
}
```

**didChangeConfiguration**
```
{
  "php": {
    "executablePath": "\/usr\/local\/opt\/php71\/bin\/php",
    "fileTypes": [
      ".php"
    ],
    "suggest": {
      "basic": true
    },
    "validate": {
      "enable": true,
      "executablePath": "\/usr\/local\/opt\/php71\/bin\/php",
      "run": "onSave"
    }
  }
}
```

felixfbecker: It's up to us to defined -  I am just saying it's a bit confusing that they are different. It would be much easier for a client if we just document please pass the `didChangeConfiguration` settings as `initializationOptions`

src/Options.php
```diff
@@ -0,0 +1,49 @@
...
+
+    /**
+     * Validate/Filter input and set options for file types
+     *
+     * @param array $fileTypes List of file types
+     */
+    public function setFileTypes(array $fileTypes)
+    {
+        $fileTypes = filter_var_array($fileTypes, FILTER_SANITIZE_STRING);
+        $fileTypes =  filter_var($fileTypes, FILTER_CALLBACK, ['options' => [$this, 'filterFileTypes']]); // validate file type format
```
haringsrob: There is one space to much here.

jens1o:
any news on this?

haringsrob:
I dont think this will be done before #357 is done, and then it might be obsolete/require to be redone

martinfojtik:
is somebody working on this?

JSteitz:
Currently no.

Need to find some time to review and update to upstream.

JSteitz:
After revisiting this PR I removed all code blocks that are not needed.
Options are currently only relevant on startup and not "on the go", 
meaning we need to restart vscode, when changes affect indexing.

My plans for now are:

- [x] Options that affect indexing require restart of vscode/languageserver
- [ ] Show notification for required restart

JSteitz:
@felixfbecker I have difficulties with race conditions. Any hints how I can make the `workspace/configuration` request before the `didChangeConfiguration` message is processed?.


Expected order:

* `initialize`
* `initialized`
* `didChangeConfiguration`

What actually happens:

* `initialize`
* `initialized` (until requesting settings)
* `didChangeConfiguration` (workspace not initialized at this point)
* `initialized` (rest)

felixfbecker:
Without fully understanding what's causing problems for you, you can save a `Promise` that you fulfill explicitely when initialization is done and `yield` in the coroutine that needs to wait for it. You can also emit it as an event on an `EventEmitter` and use a util like [`waitForEvent`](https://github.com/felixfbecker/php-language-server/blob/b4b4a2fff5904f4ce895587518642ef7c19771cc/src/utils.php#L82-L94) to get a Promise for it.

JSteitz:
Looks like I have a general problem with `initialized` and the usage of `workspace/configuration`. (https://github.com/felixfbecker/php-language-server/pull/308/commits/e317e8c743a4b71cf5a97dca23c1cf67901f6685#diff-b7156ac1be31f214d6450d4dbfa3d15fR293)

After investigating vscode-languageclient, I could see that the client is sending the reponse to my request after `workspace/didChangeConfiguration`, `textDocument/didOpen` and `textDocument/documentSymbol` are sent (last two when a php file is open at startup).
So it's not racing problem, my request is just queued after the initial requests :/

Both `workspace` and `textDocument` are not initialized at that point.
`workspace` requires a dependecny to `indexer` to allow reindexing on `didChangeConfiguratioin`.

Do we drop this part and only implement `didChangeConfiguration`?
`initializationOptions` is still available :)

JSteitz:
Current state:

- [x] ~~Options that affect indexing require restart of vscode/languageserver~~
- [x] Log notification for cancel request
- [x] Log notification for reindexing
- [x] Reindex when affected settings change
- [ ] 🏃 Handle cancel request in indexer
- [x] Start indexer after initialization
- [x] Request settings if capability is supported
- [ ] Tests
- [ ] Documentation

The cancel request for reindexing is working, but I'm not so happy with that.
For now you will get indexing messages (e.g. "Indexing project for definitions and static references"), but it will not run them.
Also the final message "X files are indexed in ..." is also wrong.

I was thinking to wrap the index in a try catch and throw an exception to abort indexing altogether,
but I don't think this is the right approach.

src/Indexer.php
```diff
@@ -196,6 +230,35 @@ public function index(): Promise
...
+     * @return Promise
+     */
+    public function cancel(): Promise
+    {
+        return coroutine(function () {
+            $this->hasCancellationSignal = true;
+
+            while ($this->isIndexing()) {
+                yield timeout();
+            }
```
felixfbecker: I'm not that happy with this "polling" solution - usually cancellation is synchronous and best-effort. I.e. the old indexing would stop _eventually_ but you can already start with the next index run, by just making sure to instantiate new Index objects (maybe even a new Indexer) so that the old run can't accidentally still write to the new index. Does that sound feasible?

src/Protocol/ConfigurationItem.php
```diff
@@ -0,0 +1,20 @@
+<?php
+
+namespace LanguageServer\Protocol;
+
+class ConfigurationItem
```
felixfbecker: A constructor would be good to reduce boilerplate

src/LanguageServer.php
```diff
@@ -289,10 +324,48 @@ public function initialize(ClientCapabilities $capabilities, string $rootPath =
...
+                return;
+            }
+
+            // request configuration if it is supported
+            // support comes with protocol version 3.6.0
+            if ($this->clientCapabilities->workspace->configuration) {
+                $configurationitem = new ConfigurationItem();
+                $configurationitem->section = 'php';
+                $configuration = yield $this->client->workspace->configuration([$configurationitem]);
+                $options = $this->mapper->map($configuration[0], new Options());
```
felixfbecker: Shouldn't this be done by the `configuration` method?

src/LanguageServer.php
```diff
@@ -289,10 +324,48 @@ public function initialize(ClientCapabilities $capabilities, string $rootPath =
...
+            }
+
+            // depending on the implementation of the client
+            // the workspace/didChangeConfiguration can be invoked before
+            // the response from the workspace/configuration request is resolved
+            if ($this->indexer->isIndexing()) {
+                return;
```
felixfbecker: And why do you need to return in that case?

src/LanguageServer.php
```diff
@@ -289,10 +324,48 @@ public function initialize(ClientCapabilities $capabilities, string $rootPath =
...
+            }
+
+            if ($options) {
+                $this->indexer->setOptions($options);
+            }
+
+            $this->indexer->index()->otherwise('\\LanguageServer\\crash');
```
felixfbecker: You should `yield` this instead and make the whole coroutine crash on error since notifications can't return error responses

src/Indexer.php
felixfbecker: The Glob library should throw an exception for this, so that should not be needed.

src/LanguageServer.php
JSteitz: https://github.com/felixfbecker/php-language-server/pull/308#issuecomment-417713457

Because when the settings are different from default, the `didChangeConfiguration` is initiating the indexer. When it is already running, we know the running indexer is using the latest settings.

src/LanguageServer.php
JSteitz: As long as we don't say, we will only use settings from the section "php", i would not do that, because I know that people will ask for including `files.exclude` setting (see https://github.com/felixfbecker/php-language-server/issues/159#issuecomment-306143750).

src/Indexer.php
JSteitz: Don't think a new Indexer instance is a good idea, they are used in 2 different places (`initialized` and `didChangeConfiguration`) and I would need a reference to both.

And I think creating new Index objects is too much work from what I see.
Are you talking about `sourceIndex` and `dependencyIndex`?

If yes, then the method would need to know about all objects that require the index objects. Or am I wrong here?

PR #315, opened 2017-02-24
show anything from a doc comment block

### LINKED ISSUE #312: Hover shows just the very first doc comment paragraph
image
image
https://github.com/slimphp/Slim/blob/3.x/Slim/App.php

jens1o:
fixes #312

### DIFF
--- a/src/DefinitionResolver.php
+++ b/src/DefinitionResolver.php
@@ -89,7 +89,7 @@ class DefinitionResolver
         } else {
             $docBlock = $node->getAttribute('docBlock');
             if ($docBlock !== null) {
-                return $docBlock->getDescription();
+                return $docBlock->getSummary() . "\n" . $docBlock->getDescription();
             }
         }
     }
--- a/src/DefinitionResolver.php
+++ b/src/DefinitionResolver.php
@@ -89,7 +89,7 @@ class DefinitionResolver
         } else {
             $docBlock = $node->getAttribute('docBlock');
             if ($docBlock !== null) {
-                return $docBlock->getSummary();
+                return $docBlock->getDescription();
             }
         }
     }
--- a/src/DefinitionResolver.php
+++ b/src/DefinitionResolver.php
@@ -93,7 +93,7 @@ class DefinitionResolver
                 // with the description
                 $description = $docBlock->getDescription()->render();
 
-                if (empty(($description))) {
+                if (empty($description)) {
                     return $docBlock->getSummary();
                 } else {
                     return $docBlock->getSummary() . "\n\n" . $description;
--- a/src/DefinitionResolver.php
+++ b/src/DefinitionResolver.php
@@ -95,9 +95,8 @@ class DefinitionResolver
 
                 if (empty($description)) {
                     return $docBlock->getSummary();
-                } else {
-                    return $docBlock->getSummary() . "\n\n" . $description;
                 }
+                return $docBlock->getSummary() . "\n\n" . $description;
             }
         }
     }
--- a/src/DefinitionResolver.php
+++ b/src/DefinitionResolver.php
@@ -89,7 +89,7 @@ class DefinitionResolver
         } else {
             $docBlock = $node->getAttribute('docBlock');
             if ($docBlock !== null) {
-                return $docBlock->getSummary() . "\n" . $docBlock->getDescription();
+                return $docBlock->getSummary() . "\n\n" . $docBlock->getDescription();
             }
         }
     }
--- a/src/DefinitionResolver.php
+++ b/src/DefinitionResolver.php
@@ -89,7 +89,11 @@ class DefinitionResolver
         } else {
             $docBlock = $node->getAttribute('docBlock');
             if ($docBlock !== null) {
-                return $docBlock->getSummary() . "\n\n" . $docBlock->getDescription();
+                if (empty($docBlock->getDescription()->render())) {
+                    return $docBlock->getSummary();
+                } else {
+                    return $docBlock->getSummary() . "\n\n" . $docBlock->getDescription();
+                }
             }
         }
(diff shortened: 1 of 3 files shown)


src/DefinitionResolver.php
```diff
@@ -89,7 +89,7 @@ public function getDocumentationFromNode(Node $node)
         } else {
             $docBlock = $node->getAttribute('docBlock');
             if ($docBlock !== null) {
-                return $docBlock->getSummary();
+                return $docBlock->getSummary() . "\n" . $docBlock->getDescription();
```
felixfbecker: I think this should be a double line break because it represents a paragraph in markdown and they are separated in the source through a double line break

jens1o: You're right, thanks. But why did the tests fail?

jens1o:
I found out why the tests fail. We should test first wether a description exists, and only then we should add the line breaks.

jens1o:
Okey, looks good now. But we need to rewrite the tests. Can I do it for you?

felixfbecker:
Of course you can

src/DefinitionResolver.php
```diff
@@ -89,7 +89,11 @@ public function getDocumentationFromNode(Node $node)
         } else {
             $docBlock = $node->getAttribute('docBlock');
             if ($docBlock !== null) {
-                return $docBlock->getSummary();
+                if (empty($docBlock->getDescription()->render())) {
```
felixfbecker: I would have assumed `getDescription()` to return `null`, is that not the case?
the getter and `render()` should only be called once.

jens1o: No. I also wondered about it, but I've tested it:

```php
<?php

require_once 'vendor/autoload.php';

$factory  = \phpDocumentor\Reflection\DocBlockFactory::createInstance();

$docComment = <<<DOCCOMMENT
/**
 * This is an example of a summary.
 */
DOCCOMMENT;

$docblock = $factory->create($docComment);

var_dump($docblock->getDescription()); // returns a Description class with `$bodyTemplate` => "" (string)
```

src/DefinitionResolver.php
```diff
@@ -89,10 +89,14 @@ public function getDocumentationFromNode(Node $node)
         } else {
             $docBlock = $node->getAttribute('docBlock');
             if ($docBlock !== null) {
-                if (empty($docBlock->getDescription()->render())) {
+                // check wether we have a description, when true, add a new paragraph
+                // with the description
+                $description = $docBlock->getDescription()->render();
+
+                if (empty(($description))) {
```
felixfbecker: double parenthesis

jens1o: oops, ran over from a previous, inline version.

jens1o:
I'd fixed some tests, but there are three, where I don't know what to do. Could you point me to the good direction?

tests/Server/TextDocument/CompletionTest.php
```diff
@@ -160,7 +160,12 @@ public function testNewInNamespace()
                 'TestClass',
                 CompletionItemKind::CLASS_,
                 null,
-                'Pariatur ut laborum tempor voluptate consequat ea deserunt.',
+                'Pariatur ut laborum tempor voluptate consequat ea deserunt.' . "\n\n" .
+            'Deserunt enim minim sunt sint ea nisi. Deserunt excepteur tempor id nostrud
+laboris commodo ad commodo velit mollit qui non officia id. Nulla duis veniam
+veniam officia deserunt et non dolore mollit ea quis eiusmod sit non. Occaecat
+consequat sunt culpa exercitation pariatur id reprehenderit nisi incididunt Lorem
+sint. Officia culpa pariatur laborum nostrud cupidatat consequat mollit.',
```
felixfbecker: please don't do un-indented multi-line strings. They break code folding in editors

jens1o: so I should qoute any line and add a newline at the end?

felixfbecker: yep

jens1o:
any news on this?

src/DefinitionResolver.php
```diff
@@ -89,7 +89,15 @@ public function getDocumentationFromNode(Node $node)
...
             $docBlock = $node->getAttribute('docBlock');
             if ($docBlock !== null) {
-                return $docBlock->getSummary();
+                // check wether we have a description, when true, add a new paragraph
+                // with the description
+                $description = $docBlock->getDescription()->render();
+
+                if (empty($description)) {
+                    return $docBlock->getSummary();
+                } else {
```
felixfbecker: Remove the `else`

tests/Server/TextDocument/HoverTest.php
```diff
@@ -22,7 +22,12 @@ public function testHoverForClassLike()
         )->wait();
         $this->assertEquals(new Hover([
             new MarkedString('php', "<?php\nclass TestClass implements \\TestInterface"),
-            'Pariatur ut laborum tempor voluptate consequat ea deserunt.'
+            'Pariatur ut laborum tempor voluptate consequat ea deserunt.' . "\n\n" .
+            'Deserunt enim minim sunt sint ea nisi. Deserunt excepteur tempor id nostrud
+laboris commodo ad commodo velit mollit qui non officia id. Nulla duis veniam
+veniam officia deserunt et non dolore mollit ea quis eiusmod sit non. Occaecat
+consequat sunt culpa exercitation pariatur id reprehenderit nisi incididunt Lorem
+sint. Officia culpa pariatur laborum nostrud cupidatat consequat mollit.'
```
felixfbecker: Fix indendation (for code folding)

jens1o:
like this?

PR #319, opened 2017-03-01
Ensure diagnostics are cleared on file deletion

mousetraps:
Previously, error diagnostics would not be cleared when a file was deleted while it was closed. This would result in lingering errors in the problems view that could only be cleared by reloading the language server. This fix addresses the issue by adding support for workspace/didChangeWatchedFiles and automatically clearing diagnostics for deleted files.

**Additional Notes / Questions**
* can add some more tests if it turns out that this is the correct approach, but wanted to get it out there for early feedback. 
* Is there a better way to test this? e.g. Currently the test is not resilient to the method name not matching the protocol name (test would succeed, but scenario would fail in real life)
* I discovered another couple issues while working through this which would be much easier to address after this fix:
  * language server only pulls in files during initialize, which means that files added from outside of the current editor instance are not being picked up. 
  * language server does not properly update definitions for deleted, changed, or moved files that are not opened in the editor.
* the change also requires a fix to vscode-php-intellisense to enable fileEvents in clientOptions. If there are options to include non *.php files in language server completions, this should be considered.
* I'm fairly certain PHP 7 optimizes empty array initializations, but need to double check on this.

### DIFF
--- a/tests/Server/Workspace/DidChangeWatchedFilesTest.php
+++ b/tests/Server/Workspace/DidChangeWatchedFilesTest.php
@@ -22,12 +22,10 @@ class DidChangeWatchedFilesTest extends ServerTestCase
         $loader = new PhpDocumentLoader(new FileSystemContentRetriever(), $projectIndex, $definitionResolver);
         $workspace = new Server\Workspace($client, $projectIndex, $dependenciesIndex, $sourceIndex, null, $loader, null);
 
-        $fileEvent = new FileEvent();
-        $fileEvent->uri = 'my uri';
-        $fileEvent->type = FileChangeType::DELETED;
+        $fileEvent = new FileEvent('my uri', FileChangeType::DELETED);
 
         $isDiagnosticsCleared = false;
-        $writer->on('message', function (Message $message) use ($fileEvent, & $isDiagnosticsCleared) {
+        $writer->on('message', function (Message $message) use ($fileEvent, &$isDiagnosticsCleared) {
             if ($message->body->method === "textDocument/publishDiagnostics") {
                 $this->assertEquals($message->body->params->uri, $fileEvent->uri);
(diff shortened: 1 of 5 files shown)


tests/Server/Workspace/DidChangeWatchedFilesTest.php
```diff
@@ -0,0 +1,43 @@
...
+    {
+        $client = new LanguageClient(new MockProtocolStream(), $writer = new MockProtocolStream());
+        $projectIndex = new ProjectIndex($sourceIndex = new Index(), $dependenciesIndex = new DependenciesIndex());
+        $definitionResolver = new DefinitionResolver($projectIndex);
+        $loader = new PhpDocumentLoader(new FileSystemContentRetriever(), $projectIndex, $definitionResolver);
+        $workspace = new Server\Workspace($client, $projectIndex, $dependenciesIndex, $sourceIndex, null, $loader, null);
+
+        $fileEvent = new FileEvent();
+        $fileEvent->uri = 'my uri';
+        $fileEvent->type = FileChangeType::DELETED;
```
felixfbecker: I think it makes sense to add a constructor to `FileEvent` to make this easier to construct

tests/Server/Workspace/DidChangeWatchedFilesTest.php
```diff
@@ -0,0 +1,43 @@
...
+
+        $isDiagnosticsCleared = false;
+        $writer->on('message', function (Message $message) use ($fileEvent, & $isDiagnosticsCleared) {
```
felixfbecker: No space after `&`

mousetraps:
Done, thanks for the feedback! Also made the corresponding PR at https://github.com/felixfbecker/vscode-php-intellisense/pull/83. And with that, I'm off to bed because it's 2am here. 💤

PR #320, opened 2017-03-02
Enforce memory limit in phpunit.xml

mousetraps:
This will help highlight memory regressions, make it easier for newcomers to get started with the codebase w/o editing php.ini defaults (128M), and also keep things consistent between local and travis runs.


mousetraps:
Test failure is unrelated to this change, and should pass after #322 is merged.

PR #322, opened 2017-03-02
Fix formatting breakage

### LINKED ISSUE #321: Formatting breaks after most recent PHPCS update
Repro steps:
1. grab the latest source from master
2. run `composer update` to update dependencies
3. formatting operations fail with:

PHP Fatal error:  main(): Failed opening required 'C:\src\php\php-language-server\src/../vendor/squizlabs/php_codesniffer/Config.php' (include_path='.;C:\php\pear') in C:\src\php\php-language-server\src\Formatter.php on line 6

The issue is also causing the travis build to fail. Working on a fix.

mousetraps:
PHP_CodeSniffer 3.0 RC4 introduces a breaking change that removes PHPCS from the composer autoloader. This fix adds a bootstrapping step so that we can invoke the autoloader automatically. Tests should pass after this.

Also see 
* https://github.com/squizlabs/PHP_CodeSniffer/commit/566a0cdf6cdb8fbc5a5a248cdf23f7d77a968de7
* https://github.com/squizlabs/PHP_CodeSniffer/issues/1333

close #321


src/bootstrap.php
```diff
@@ -0,0 +1,5 @@
+<?php
+
+// PHP_CodeSniffer does not define an autoloader in its composer.json,
+// so we must invoke it ourselves.
+require __DIR__ . "/../vendor/squizlabs/php_codesniffer/autoload.php";
```
staabm: doesn't it miss to include the "regular" `vendor/autoload.php` now?

felixfbecker: Yeah, I'm a bit worried we are coupling ourselves to the CodeSniffer autoloader here. So far, I only had problems with it and passing CodeSniffer a string to format while we already have an AST for the document is less than ideal

mousetraps: The other option is just tieing to the RC3 release, which is not broken.

felixfbecker: There was at the time a reason why I had to depend on the unstable version, it had an important bug fix - maybe it's fixed now in the most recent RC, might be worth to try out

felixfbecker: Wait, just saw that I do not depend on the unstable version anymore:
```json
        "squizlabs/php_codesniffer" : "^3.0",
```
I wonder how the breaking change got in there then, as `preferStable` is set?
Anyway, if it fixes it and tests pass I would prefer to lock the most recent version that works

mousetraps: Because there aren't any stable versions for 3.0 yet (2.x is the current stable version)

mousetraps:
Closing in favor of #324

PR #328, opened 2017-03-03
Fix crash when returning null instead of a Type when trying to infer node type

Alenore:
Returns a Mixed instead of Null_, as the return type is pretty much unknown.

### DIFF
--- a/src/DefinitionResolver.php
+++ b/src/DefinitionResolver.php
@@ -813,7 +813,9 @@ class DefinitionResolver
             // Unknown
             return new Types\Mixed;
         }
-        return null;
+        // Return Types\Mixed in case nothing was found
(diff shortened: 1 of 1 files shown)


felixfbecker:
The return value has a semantic meaning

> Returns null if the node does not have a type.

Only very few nodes have a type.

drKnoxy:
just wanted to say thanks @Alenore my plugin was busted after upgrading from 0.0.12 -> 1.1.3 and this fixed it.

felixfbecker:
@Alenore Are you planning to look into the cause of the issue?

Alenore:
I was especially busy recently but it's still on my todo list, as it directly affects my workflow :)

Alenore:
Apparently was resolved in PR #340

Issue #326, opened 2017-03-02
Connection to server got closed.


designcraft:
Hi,

VS Code: 1.10.1
Node: 6.5.0
PHP extension pack: 1.0.0
Php intellisense: 1.1.3

I have this problem since installed the extension, server got closed and will not be restarted after 5 crash.
Console log:

extensionHost.ts:282 [Extension Host] TypeError: Return value of LanguageServer\DefinitionResolver::resolveExpressionNodeToType() must implement interface phpDocumentor\Reflection\Type, null returned in C:\Users\Varga S�ndor\.vscode\extensions\felixfbecker.php-intellisense-1.1.3\vendor\felixfbecker\language-server\src\DefinitionResolver.php:445
Stack trace: ....

Alenore:
I can provide a bit more information about this, and probably a mean to reproduce the problem : it happens on 3 different projects when trying to parse acf.php from the ACF WordPress plugin.


As a workaround, i've changed the `return null;` at line 816 in the file DefinitionResolver.php (function getTypeFromNode) to `return new Types\Null_;`

designcraft:
Hello Alenore,
Where can I find the DefinitionResolver.php file? I searched but not found.

designcraft:
I just copy the custom post type code from ACF, and CPT ui and pasted it to the functions.php than removed these plugins from WP. Finally the Intellisense extension works, so thank you Alenore!

Alenore:
@designcraft I wouldn't advice you to do that, as it's going to pain to support ACF updates over the time.
As I know you're on Win10, the file is located here : 
%USERPROFILE%\.vscode\extensions\felixfbecker.php-intellisense-1.1.3\vendor\felixfbecker\language-server
(Switch directory version depending on your current installed extension ;))

felixfbecker:
@Alenore @designcraft If you want painless future updates and have time to look at this, why not just do a PR? :)

Alenore:
@felixfbecker The main reason why I didn't do a PR is, my fix is an ugly one :p It doesn't really fix the source of the problem (why the function returns null in the first place, instead of correctly matching the type), just returns a Types\Null_ when it could possibly be something else. 
If you're fine with this temporary workaround, I don't mind doing a PR :)

felixfbecker:
It just seems like you all are willing to invest so much time in the workaround, so just wanted to remind that this is a community project and anyone can find the real fix and submit it ;)

laperon:
[...]

PR #333, opened 2017-03-16
Update DefinitionResolver.php

Sensirex:
from issue
https://github.com/felixfbecker/php-language-server/issues/326


felixfbecker:
As I already said before, this is not the fix. The bug is that `type` gets _set_ to `null`, not that a `null` value isn't handled.

Issue #311, opened 2017-02-20
Crash because Definition->type is null for a function


lumpysimon:
Indexing a WordPress site, PHP7.1.1 on OSX, after about 15 mins of indexing I'm getting the `The PHP Language Server server crashed 5 times in the last 3 minutes. The server will not be restarted` error. Here's the stack trace:

````
extensionHost.ts:282 [Extension Host] TypeError: Return value of LanguageServer\DefinitionResolver::resolveExpressionNodeToType() must implement interface phpDocumentor\Reflection\Type, null returned in /Users/myuserhere/.vscode/extensions/felixfbecker.php-intellisense-1.1.2/vendor/felixfbecker/language-server/src/DefinitionResolver.php:444
Stack trace:
#0 /Users/myuserhere/.vscode/extensions/felixfbecker.php-intellisense-1.1.2/vendor/felixfbecker/language-server/src/DefinitionResolver.php(238): LanguageServer\DefinitionResolver->resolveExpressionNodeToType(Object(PhpParser\Node\Expr\FuncCall))
#1 /Users/myuserhere/.vscode/extensions/felixfbecker.php-intellisense-1.1.2/vendor/felixfbecker/language-server/src/NodeVisitor/ReferencesCollector.php(38): LanguageServer\DefinitionResolver->resolveReferenceNodeToFqn(Object(PhpParser\Node\Expr\MethodCall))
#2 /Users/myuserhere/.vscode/extensions/felixfbecker.php-intellisense-1.1.2/vendor/nikic/php-parser/lib/PhpParser/NodeTraverser.php(108): LanguageServer\NodeVisitor\ReferencesCollector->enterNode(Object(PhpParser\Node\Expr\MethodCall))
#3 /Users/myuserhere/.vscode/extensions/felixfbecker.php-intellisense-1.1.2/vendor/nikic/php-parser/lib/PhpParser/NodeTraverser.php(171): PhpParser\NodeTraverser->traverseNode(Object(PhpParser\Node\Expr\Assign))
#4 /Users/myuserhere/.vscode/extensions/felixfbecker.php-intellisense-1.1.2/vendor/nikic/php-parser/lib/PhpParser/NodeTraverser.php(101): PhpParser\NodeTraverser->traverseArray(Array)
[...]

Issue #325, opened 2017-03-02
The PHP Language Server server crashed 5 times in the last 3 minutes. The server will not be restarted.


wondergryphon:
Hiya

Found a few issues with the above error message, but none of the resolutions of those issues have helped me.

**Details:**
macOS Sierra
php v7
node v6.9.1
npm -v4.3.0

**Stack Trace from VSCode**
```
//Applications/Visual Studio Code.app/Contents/Resources/app/out/vs/workbench/electron-browser/workbench.main.js:136 [Extension Host] TypeError: Return value of LanguageServer\DefinitionResolver::resolveExpressionNodeToType() must implement interface phpDocumentor\Reflection\Type, null returned in /Users/markus/.vscode/extensions/felixfbecker.php-intellisense-1.1.3/vendor/felixfbecker/language-server/src/DefinitionResolver.php:445
Stack trace:
#0 /Users/markus/.vscode/extensions/felixfbecker.php-intellisense-1.1.3/vendor/felixfbecker/language-server/src/DefinitionResolver.php(239): LanguageServer\DefinitionResolver->resolveExpressionNodeToType(Object(PhpParser\Node\Expr\FuncCall))
#1 /Users/markus/.vscode/extensions/felixfbecker.php-intellisense-1.1.3/vendor/felixfbecker/language-server/src/NodeVisitor/ReferencesCollector.php(38): LanguageServer\DefinitionResolver->resolveReferenceNodeToFqn(Object(PhpParser\Node\Expr\PropertyFetch))
#2 /Users/markus/.vscode/extensions/felixfbecker.php-intellisense-1.1.3/vendor/nikic/php-parser/lib/PhpParser/NodeTraverser.php(108): LanguageServer\NodeVisitor\ReferencesCollector->enterNode(Object(PhpParser\Node\Expr\PropertyFetch))
#3 /Users/markus/.vscode/extensions/felixfbecker.php-intellisense-1.1.3/vendor/nikic/php-parser/lib/PhpParser/NodeTraverser.php(171): PhpParser\NodeTraverser->traverseNode(Object(PhpParser\Node\Expr\Assign))
#4 /Users/markus/.vscode/extensions/felixfbecker.php-intellisense-1.1.3/vendor/nikic/php-parser/lib/PhpParser/NodeTraverser.php(101): PhpParser\NodeTraverser->traverseArray(Array)
[...]

PR #340, opened 2017-04-01
Fix missing '()' for function definition

Talv:
Should fix:
#311
#325
#326

### DIFF
--- a/src/DefinitionResolver.php
+++ b/src/DefinitionResolver.php
@@ -446,7 +446,7 @@ class DefinitionResolver
                 // Cannot get type for dynamic function call
                 return new Types\Mixed;
             }
-            $fqn = (string)($expr->getAttribute('namespacedName') ?? $expr->name);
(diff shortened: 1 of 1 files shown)


src/DefinitionResolver.php
```diff
@@ -446,7 +446,7 @@ public function resolveExpressionNodeToType(Node\Expr $expr): Type
                 // Cannot get type for dynamic function call
                 return new Types\Mixed;
             }
-            $fqn = (string)($expr->getAttribute('namespacedName') ?? $expr->name);
+            $fqn = (string)($expr->getAttribute('namespacedName') ?? $expr->name . '()');
```
felixfbecker: Pretty sure `namespacedName` would need to get the `()` too, so you have to move the parenthesis

Talv: Yeah, you're right. I have made wrong assumption as to what ```namespacedName``` actually contains.

felixfbecker:
Thanks for getting to the bottom of this!

PR #347, opened 2017-04-10
Support constants

### LINKED ISSUE #48: Support define() constants

jens1o:
Yes, I used the master branch. So I have less problems 😉 

resolves #84 and fixes #48


jens1o:
Seems to work(excluding the code style). Did I missed something @felixfbecker?

tests/Server/TextDocument/HoverTest.php
```diff
@@ -156,6 +156,21 @@ public function testHoverForConstant()
         ], $reference->range), $result);
     }
 
+    public function testHoverForGlobalConstant()
+    {
+        // print TEST_PROPERTY ? 'true' : 'false';
+        // Get hover for TEST_PROPERTY
+        $reference = $this->getReferenceLocations('TEST_PROPERTY')[0];
```
felixfbecker: Please name it `TEST_DEFINE_CONSTANT` or something like that, `TEST_PROPERTY` is confusing

jens1o:
any news on this?

tests/Server/Workspace/SymbolTest.php
```diff
@@ -58,9 +58,10 @@ public function testEmptyQueryReturnsAllSymbols()
             new SymbolInformation('TestInterface',      SymbolKind::INTERFACE,   $this->getDefinitionLocation('TestInterface'),                                ''),
             new SymbolInformation('test_function',      SymbolKind::FUNCTION,    $this->getDefinitionLocation('test_function()'),                              ''),
             new SymbolInformation('ChildClass',         SymbolKind::CLASS_,      $this->getDefinitionLocation('ChildClass'),                                   ''),
+            new SymbolInformation('define',             SymbolKind::CONSTANT,    $this->getDefinitionLocation('TEST_PROPERTY'),                                ''),
```
felixfbecker: If this is the defined constant, shouldn't the symbol name be `TEST_PROPERTY` (or `TEST_DEFINE_CONSTANT`), not `define`?

jens1o: No, because it's the name of it. The value is the `TEST_PROPERTY`. But I can change that if you want.

felixfbecker: What do you mean with value? A SymbolInformation has a name, and the name of the defined constant is TEST_PROPERTY. What value?

jens1o: ah, that's a bug. The name had been overwritten...

src/Protocol/SymbolInformation.php
```diff
@@ -91,6 +90,14 @@ public static function fromNode(Node $node, string $fqn = null)
...
         if ($node instanceof Node\Name) {
             $symbol->name = (string)$node;
+        } else if(
+            $node instanceof Node\Expr\FuncCall
+            && $node->name instanceof Node\Name
+            && strtolower((string)$node->name) === 'define'
+            && isset($node->args[0])
+            && $node->args[0]->value instanceof Node\Scalar\String_
+        ) {
+            $symbol->name = (string)$node->args[0]->value->value;
```
jens1o: I'd added this because otherwise it would get renamed immediatly in line 98(105)

felixfbecker: l98 only sets `containerName`?

jens1o:
Seems to work, did I missed something else?

src/Protocol/SymbolInformation.php
jens1o: In that line?
```php
 } else if (isset($node->name)) {
             $symbol->name = (string)$node->name;
} 
```

I'd debugged a bit and figured out that without this method it will be set to `define`(which is the name of the node). But we want the value as a name, so that's why.

felixfbecker: You don't wanna check the exact same condition twice.

jens1o: I need to. Otherwise it would return null in this line: https://github.com/jens1o/php-language-server/blob/7380acc49fbacb4d5a6b3a7bf76d6dd55eed2043/src/Protocol/SymbolInformation.php#L88-L90 

And otherwise it would get resetted when I don't check it twice.

felixfbecker: yeah, I see, but there are better options like setting the name early and checking later if the name is already set.

jens1o: okay. I'll do that.

jens1o:
Sorry for all these commits with the code style... I know my codestyle pretty well and forget to switch to other projects very often...

src/Protocol/SymbolInformation.php
```diff
@@ -87,10 +99,13 @@ public static function fromNode(Node $node, string $fqn = null)
         } else if ($node instanceof Node\Expr\ClosureUse) {
             $symbol->name = $node->var;
         } else if (isset($node->name)) {
```
felixfbecker: this is equivalent to `else if (isset($node->name) && !isset($symbol->name)`

jens1o: Okay, I would do this like I did, for readability, but I changed it.

jens1o: Don't seems to be identical...

Issue #350, opened 2017-04-11
Support type hinting for variables


belazor:
If I have a variable that's type hinted like so:

<img width="830" alt="screenshot 2017-04-11 at 8 41 54 pm" src="https://cloud.githubusercontent.com/assets/9424286/24927586/6ce33d82-1ef7-11e7-8a66-eb9d98ee8d18.png">

I expect to see the method definitions for the methods I call on the $emailChange variable. This does not happen:

<img width="831" alt="screenshot 2017-04-11 at 8 43 42 pm" src="https://cloud.githubusercontent.com/assets/9424286/24927640/a301e882-1ef7-11e7-93be-6926fd6cc8cf.png">

This type hinting syntax is allegedly supported by phpStorm, and it would be wonderful to have support for this in VSCode via your IntelliSense plugin as well.

Thank you in advance :)

felixfbecker:
Definitely, there is already handling for `@params` tags, so it would be trivial to add `@var` tags handling if someone wants to take a shot at it.

belazor:
Would you be able to give me a hint as to where to implement this? I could probably handle it if I had an idea of where to look :)

felixfbecker:
In this function: https://github.com/felixfbecker/php-language-server/blob/master/src/DefinitionResolver.php#L726

belazor:
I guess I'm not advanced enough a PHP coder, I can't work out how to actually test any changes I make. Trying to run the web server just produces a bunch of Undefined Index errors related to headers in the ProtocolStreamer.php file.

I'm seeing around line 798 that variables should have @var tags supported, unless I'm just horribly misreading the code? Does that make the fact that this isn't working a bug?

If it isn't a bug and I'm just failing at reading your code, would you please consider adding it yourself, as you're more intimately familiar with the code? I would really appreciate it!

Sorry I couldn't be of more help :(

felixfbecker:
The language server has nothing to do with a web server or headers ;)
See the README for how to run tests.
`@var` tags are read for properties, not variables.

belazor:
I still don't understand, sorry. 

The only mention of testing that I could find in the README file was a reference to unit testing. I don't have much experience with unit testing, so me running a command that says "149 / 149 (100%)" doesn't do much when I'm struggling with actually working out how to read the stubs file to find out if any changes are working. 

I guess I just have to hope someone is kind enough to implement this for me :)

jens1o:
[...]

PR #352, opened 2017-04-12
Support type hinting

jens1o:
This pr does include the pr #374, because the pr was created with the master branch. So I would like that the "define-constants" pr will be merged first.


jens1o:
When fine, I'll create tests. But I'm unsure wether the code is okay like this.

swiffer:
@felixfbecker is this still working, would be great to have this merged?

should solve issue #350 right?

felixfbecker:
This is based on an old version of the langserver that used a different parser, see the merge conflicts. It's also missing some tests.

jens1o:
Okay, ehm, I will watch out when I got more time and then refile the pull request.

Issue #286, opened 2017-02-07
TypeError: ... in DefinitionResolver.php:444


sebet:
This issue is similar to #252 but it happens inside the `if ($expr instanceof Node\Expr\FuncCall) {` logic.

This is the error I get in _DefinitionResolver.php_ when the server is parsing this [file](https://gist.github.com/sebet/a13f0fda9e1898b5604dfec83401c674):

```extensionHost.ts:282 [Extension Host] TypeError: Return value of LanguageServer\DefinitionResolver::resolveExpressionNodeToType() must implement interface phpDocumentor\Reflection\Type, null returned in /Users/bcarreco/.vscode/extensions/felixfbecker.php-intellisense-1.1.1/vendor/felixfbecker/language-server/src/DefinitionResolver.php:444```

I was able to fix it by adding this code to that line:

```
            if (strtolower((string)$def) === 'null') {
                return new Types\Null_;
            }
```

final changed code:
```
        if ($expr instanceof Node\Expr\FuncCall) {
            // Find the function definition
            if ($expr->name instanceof Node\Expr) {
                // Cannot get type for dynamic function call
                return new Types\Mixed;
            }
            $fqn = (string)($expr->getAttribute('namespacedName') ?? $expr->name);
            $def = $this->index->getDefinition($fqn, true);
            if (strtolower((string)$def) === 'null') {
                return new Types\Null_;
            }
            if ($def !== null) {
                return $def->type;
            }
        }
```

Not sure that's the right approach but it fixed the problem.

carmanchris31:
I've been experiencing this same issue.

```TypeError: Return value of LanguageServer\DefinitionResolver::resolveExpressionNodeToType() must implement interface phpDocumentor\Reflection\Type, null returned in /Users/christopher_carman/.vscode/extensions/felixfbecker.php-intellisense-1.1.3/vendor/felixfbecker/language-server/src/DefinitionResolver.php:445
Stack trace:
#0 /Users/christopher_carman/.vscode/extensions/felixfbecker.php-intellisense-1.1.3/vendor/felixfbecker/language-server/src/DefinitionResolver.php(239): LanguageServer\DefinitionResolver->resolveExpressionNodeToType(Object(PhpParser\Node\Expr\FuncCall))
[...]

PR #362, opened 2017-04-22
Fix for #286

belazor:


felixfbecker:
As I said numerous times, the issue is not that `null` is not handled, the issue is that the property gets set to `null` in the first place.

PR #367, opened 2017-04-26
Cache source packages when reference/revision is available

### LINKED ISSUE #366: Cache Key is not set when dependent repository targets a specific commit
We use a few dependencies in our composer.json that target a specific commit (e.g. `dev-master#4b7c20db3c72a859df049701db43659187bc4573`). Every time VSCode starts, these dependencies are reparsed, even if no file has changed since the last time VSCode has started. Within the log output, this is seen:
```
[Warn  - 9:35:35 AM] Could not compute cache key for phing/phing
```

It looks like the cache key is decided in [src/Indexer.php](https://github.com/felixfbecker/php-language-server/blob/d5c54ac30f9c41bb6cd37de8a9476117b7a2a1e7/src/Indexer.php#L151). When a version isn't used, could you cache off of the commit ID (if available)?

adduc:
Fixes #366

### DIFF
--- a/src/Indexer.php
+++ b/src/Indexer.php
@@ -149,11 +149,19 @@ class Indexer
                 $cacheKey = null;
                 $index = null;
                 foreach (array_merge($this->composerLock->packages, $this->composerLock->{'packages-dev'}) as $package) {
-                    // Check if package name matches and version is absolute
-                    // Dynamic constraints are not cached, because they can change every time
+                    // Check if package can be cached.
                     $packageVersion = ltrim($package->version, 'v');
+                    // If package is anchored to a version
                     if ($package->name === $packageName && strpos($packageVersion, 'dev') === false) {
                         $packageKey = $packageName . ':' . $packageVersion;
+                        $cacheKey = self::CACHE_VERSION . ':' . $packageKey;
+                        // Check cache
+                        $index = yield $this->cache->get($cacheKey);
+                        break;
+
+                    // If package is checked out
+                    } elseif ($package->name === $packageName && isset($package->source->reference)) {
+                        $packageKey = $packageName . ':' . $package->source->reference;
                         $cacheKey = self::CACHE_VERSION . ':' . $packageKey;
                         // Check cache
                         $index = yield $this->cache->get($cacheKey);


adduc:
Regarding testing for this change, I'm still wrapping my head around the code. I think it can be accomplished by including a composer.json and composer.lock within fixtures to mock a faux install. I'm going to look into this further.

src/Indexer.php
```diff
@@ -149,15 +149,23 @@ public function index(): Promise
                 $cacheKey = null;
                 $index = null;
                 foreach (array_merge($this->composerLock->packages, $this->composerLock->{'packages-dev'}) as $package) {
-                    // Check if package name matches and version is absolute
-                    // Dynamic constraints are not cached, because they can change every time
+                    // Check if package can be cached.
                     $packageVersion = ltrim($package->version, 'v');
+                    // If package is anchored to a version
                     if ($package->name === $packageName && strpos($packageVersion, 'dev') === false) {
```
felixfbecker: package name is checked in both conditions

src/Indexer.php
```diff
@@ -149,15 +149,23 @@ public function index(): Promise
...
                         $packageKey = $packageName . ':' . $packageVersion;
                         $cacheKey = self::CACHE_VERSION . ':' . $packageKey;
                         // Check cache
                         $index = yield $this->cache->get($cacheKey);
                         break;
+
+                    // If package is checked out
+                    } elseif ($package->name === $packageName && isset($package->source->reference)) {
```
felixfbecker: space between `else if`

felixfbecker:
LGTM overall.
Testing with a fake composer.json and composer.lock sounds like a good approach. Ideally this would be a unit test for the Indexer that asserts a stubbed Cache is called

adduc:
I've made changes based off your feedback, and added tests for the two current package caching strategies.

I think this is ready for another review.

felixfbecker:
Could you do the tests as a unit test for the Indexer?

adduc:
I'd be willing to give it a shot.

umpirsky:
Any updates on this?

PR #375, opened 2017-05-10
Perf hack

lialan:
2 things changed essentially:

1. make the global indexing part synchronized instead of asynced. This will add a few seconds to the total time but frees us from the need to wait for `definition-added` event to complete.
2. avoid re-opening/re-loading a same document's content in definition queries. We could do better with memory usage, but I think at this moment it works for the purpose.


lialan:
Sorry, wrong target ...

felixfbecker:
Makes me wonder, what was the target?

jens1o:
this: https://github.com/lambdalab/php-language-server/pull/1

PR #391, opened 2017-06-09
improve abstract detection

jens1o:


jens1o:
I decided to not use the `getText()` function, because they are only 2 possibilities and it would slow down the process...

src/DefinitionResolver.php
```diff
@@ -1190,6 +1194,11 @@ private function tryGetDocBlockTagForParameter($docBlock, $variableName)
             if ($tag->getVariableName() === \ltrim($variableName, "$")) {
                 return $tag;
             }
+        } else if ($node instanceof Node\Expr\FuncCall && $node->name instanceof Node\Name && strtolower((string)$node->name) === 'define') {
+            if (!isset($node->args[0]) || !($node->args[0]->value instanceof Node\Scalar\String_)) {
+                return null;
+            }
+            return (string)$node->args[0]->value->value;
```
felixfbecker: bad merge?

src/DefinitionResolver.php
```diff
@@ -174,7 +174,11 @@ public function createDefinitionFromNode(Node $node, string $fqn = null): Defini
         $def->fqn = $fqn;
 
         // Determines whether the suggestion will show after "new"
-        $def->canBeInstantiated = $node instanceof Node\Statement\ClassDeclaration;
+        $def->canBeInstantiated = (
+            $node instanceof Node\Statement\ClassDeclaration &&
+            // use text's length for speedup: |abstract| = 8; |final| = 5
+            ($node->abstractOrFinalModifier === null || $node->abstractOrFinalModifier->length !== 8)
```
felixfbecker: that doesn't look faster than comparing it to `TokenKind::AbstractKeyword`

jens1o: Good catch, I didn't thought about this.

src/DefinitionResolver.php
```diff
@@ -174,7 +174,11 @@ public function createDefinitionFromNode(Node $node, string $fqn = null): Defini
         $def->fqn = $fqn;
 
         // Determines whether the suggestion will show after "new"
-        $def->canBeInstantiated = $node instanceof Node\Statement\ClassDeclaration;
+        $def->canBeInstantiated = (
+            $node instanceof Node\Statement\ClassDeclaration &&
+            // check whether it is not an abstract class
+            ($node->abstractOrFinalModifier === null || $node->abstractOrFinalModifier->kind !== PhpParser\TokenKind::AbstractKeyword)
```
felixfbecker: You can remove the check for `null`

jens1o: No, I can't. Because when there is no modifier, it returns null.

jens1o: image

felixfbecker: In which case `$node->abstractOrFinalModifier->kind !== PhpParser\TokenKind::AbstractKeyword` is `true`

jens1o: That will result in a `access to undefined property 'kind'`?

felixfbecker: sorry, I thought `$kind` was `null`.

PR #392, opened 2017-06-09
Fix 'find references' for unused symbols

MadHed:
This fixes a bug where trying to execute `find references` on an unused symbol would cause the server to never return an answer, causing an infinite `waiting` in the references window.

The root of the problem is that `Promise\all` never fulfills itself when an empty array is passed.
I am not sure if this is expected behavior or a bug in the `sabre/event` lib...

### DIFF
--- a/fixtures/global_symbols.php
+++ b/fixtures/global_symbols.php
@@ -105,3 +105,15 @@ class ChildClass extends TestClass {}
 define('TEST_DEFINE_CONSTANT', false);
 
 print TEST_DEFINE_CONSTANT ? 'true' : 'false';
+
+/**
+ * Neither this class nor its members are referenced anywhere
+ */
+class UnusedClass
+{
+    public $unusedProperty;
+
+    public function unusedMethod()
+    {        
+    }
+}
--- a/src/Server/TextDocument.php
+++ b/src/Server/TextDocument.php
@@ -221,15 +221,18 @@ class TextDocument
                         return [];
                     }
                 }
-                $refDocuments = yield Promise\all(array_map(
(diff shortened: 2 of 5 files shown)


felixfbecker:
Definitely a bug in sabre/event. Example in JS:
image

felixfbecker:
Could you open an issue for this?

MadHed:
Sure, will do

felixfbecker:
tests fail

MadHed:
@see https://github.com/fruux/sabre-event/pull/50

Tests failed because I did not realize all tests in `GlobalTest` are also executed as `NamespacedTest`

staabm:
fix for sabre/event included in https://github.com/fruux/sabre-event/releases/tag/5.0.2

felixfbecker:
Oh, great! Could you change this PR to only include the tests then, and bump the version?

MadHed:
Yup doing that right now

MadHed:
I only removed the workaround code.
The dependency was already set to ^5.0, so 5.0.2 gets used.

Waiting for the tests to finish... I saw lots of failing tests on windows because of directory separator differences

felixfbecker:
Oh, maybe it's time to add Windows tests.

If you wanna do a PR, this is how I did it for the debugger:
https://github.com/felixfbecker/vscode-php-debug/blob/master/appveyor.yml
https://github.com/felixfbecker/vscode-php-debug/blob/master/appveyor-install.ps1

Just need to simplify it to exclude XDebug and Node

MadHed:
The failures stem from the new validation tests where a forward slash is hardcoded in the **expected** json files. I will have a look and see what I can do.

PR #398, opened 2017-06-11
install as a dev-dependency

staabm:
I am not totalle sure I understood the "install" chapter. please give me hints in case I missunderstood it and we should improve the docs in a different direction.

As I read this paragraphs the language-server is meant to be installed in every php project I develop locally. therefore I suppose to instruct the user to install it as a dev-dependency.

In case this paragraphs meaning is that you have to install the language server once locally and use this single instance for all php projects, I would recommend to reword this section.

thx for this great tool!

### DIFF
--- a/README.md
+++ b/README.md
@@ -130,7 +130,7 @@ All classes are considered internal and are not subject to semver.
 The recommended installation method is through [Composer](https://getcomposer.org/).
 Simply run
 
-    composer require felixfbecker/language-server
+    composer require-dev felixfbecker/language-server
 
 and you will get the latest stable release and all dependencies.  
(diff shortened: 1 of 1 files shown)


felixfbecker:
If you are an author of an editor plugin, I would certainly not recommend installing as a dev dependency. See https://github.com/felixfbecker/vscode-php-intellisense/blob/master/composer.json as an example. You usually don't have to install it for every of your projects, you need it integrated in your editor.

staabm:
so this means I just have to install the extension from the VS Code marketplace, nothing more?

felixfbecker:
If you are using VS Code, yes.

PR #405, opened 2017-06-13
Fix error getting completions for 'new static' type

roblourens:
`classTypeDesignator` can now also be a StaticKeyword, after https://github.com/Microsoft/tolerant-php-parser/pull/149


src/DefinitionResolver.php
```diff
@@ -896,11 +896,12 @@ public function resolveClassNameToType($class): Type
...
-        $className = (string)$class->getResolvedName();
-
-        if ($className === 'static') {
+        if ($class instanceof PhpParser\Token && $class->kind === PhpParser\TokenKind::StaticKeyword) {
+            // `new static`
             return new Types\Static_;
         }
+        $className = (string)$class->getResolvedName();
+
         if ($className === 'self' || $className === 'parent') {
```
felixfbecker: can these be checked through `TokenKind` too?

roblourens: There aren't TokenKinds for self and parent. I don't know why. They are parsed as QualifiedNames.

Issue #407, opened 2017-06-14
Method call whose target is field in parent class couldn't be resolved.


zfy0701:
Take following code as example, it looks `$this->f->foo();` works in base class, but doesn't work in sub class.

It looks like the problem is we alway try to construct a fqn before query, so for second one, we might have constructed a bad fqn in the first place.  Is this a intended behavior?

```php
<?php  if ( ! defined('BASEPATH')) exit('No direct script access allowed');

Class C {
	public function foo() {

	}
}

class A {
	/**
	* @var C
	*/
	public $f;

	public function __construct($c) {
		$this->f = $c;
	}

	public function test() {
		$this->f->foo(); // foo can be resolve
	}
}

class B extends A {
	public function main() {
		$this->f->foo(); // foo couldn't be resolve
	}
}
```

felixfbecker:
What exactly doesn't work? Completion or go to definition?

zfy0701:
go to defintion, you could paste this code to vs code, and try to click the two "foo" reference, the first one could correctly jump to definition but the second one can't.

autoferrit:
Same happens in my case. Example

```php
class Document {
    protected static something($param)
    {
        return array(...);
    }
}

class Pdf extends Document {
    protected static something($param)
    {
        $param[] = 'newitem';
        $parent = parent::something($param);
        return $parent;
    }
}

```

for me, when I go to  definition on `parent::something(...)` it does not know what it is. even clicking just on the `parent` part does not go to the parent class.

felixfbecker:
@autoferrit that likely is a different issue related to the `parent` keyword.

PR #408, opened 2017-06-14
Try to lookup base class member while resolveExpressionNodeToType

zfy0701:
Is this something you would do to solve 
https://github.com/felixfbecker/php-language-server/issues/407

### DIFF
--- a/src/DefinitionResolver.php
+++ b/src/DefinitionResolver.php
@@ -656,13 +656,31 @@ class DefinitionResolver
                 } else {
                     $classFqn = substr((string)$t->getFqsen(), 1);
                 }
-                $fqn = $classFqn . '->' . $expr->memberName->getText($expr->getFileContents());
-                if ($expr->parent instanceof Node\Expression\CallExpression) {
-                    $fqn .= '()';
+
+                $memberSuffix = '->' . $expr->memberName->getText($expr->getFileContents());
+                if ($expr instanceof Node\Expression\CallExpression) {
+                    $memberSuffix .= '()';
                 }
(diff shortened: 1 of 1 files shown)


src/DefinitionResolver.php
```diff
@@ -656,13 +656,31 @@ public function resolveExpressionNodeToType($expr)
...
+                if ($expr instanceof Node\Expression\CallExpression) {
+                    $memberSuffix .= '()';
                 }
-                $def = $this->index->getDefinition($fqn);
-                if ($def !== null) {
-                    return $def->type;
+
+                // Find the right class that implements the member
+                $implementorFqns = [$classFqn];
+                while ($implementorFqn = array_shift($implementorFqns)) {
```
roblourens: Looks like this block is copied from `resolveMemberAccessExpressionNodeToFqn`, maybe there's some way the code can be reused instead of duplicated. Also the comment below about FQN now doesn't match the code.

vakata:
Correct me if I am wrong but I believe this PR and #421 do pretty much the same thing. 

The differences are:
1) in the implementation of parent gathering - #421 uses a generator closure and #408 uses an array
2) ```this``` handling - #421 handles the This type to allow for fluent interfaces
3) #421 has tests (which still pass of this moment)

We should probably leave a single PR open as they do the same and merging both will not be possible. We could also mix and match the two approaches - add the This type handling and tests to #408 or adopt the array approach in #421 (if the maintainers believe it is better than the generator closure)?

zfy0701:
sounds good so I'll just close this

Issue #380, opened 2017-06-01
No autocompletion inside namespace


ewwink:
but hover is work, example code
```php
<?php
namespace mySpace;

Class Something
{
  public function someMethod()
  {
     str_ 
    // No suggestions
  }
}
```

PR #412, opened 2017-06-16
Improve CompletionProvider

felixfbecker:
- Better performance
  This moves stuff out of the `foreach` loop that doesn't have to be in the loop
- More documentation, clearer variable names
- Bring implementation closer to pre-tolerant parser implementation (for example, now always suggests aliases)
- Add `roamed` field to Definition for global namespace fallback for #380

/cc @roblourens

### DIFF
--- a/src/CompletionProvider.php
+++ b/src/CompletionProvider.php
@@ -128,6 +128,7 @@ class CompletionProvider
         // This can be made much more performant if the tree follows specific invariants.
         $node = $doc->getNodeAtPosition($pos);
 
+        // Get the node at the position under the cursor
         $offset = $node === null ? -1 : $pos->toOffset($node->getFileContents());
         if (
             $node !== null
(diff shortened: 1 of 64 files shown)


felixfbecker:
@roblourens I would merge this and #413 and then release if you have no objections

roblourens:
Cool. I'm looking through this one now.

roblourens:
LGTM!

src/CompletionProvider.php
```diff
@@ -178,138 +188,189 @@ public function provideCompletion(PhpDocument $doc, Position $pos): CompletionLi
...
+                : $nameNode->getText($node->getFileContents());
+            $prefixLen = strlen($prefix);
+
+            /** Whether the prefix is qualified (contains at least one backslash) */
+            $isQualified = $nameNode instanceof Node\QualifiedName && $nameNode->isQualifiedName();
+
+            /** Whether the prefix is fully qualified (begins with a backslash) */
+            $isFullyQualified = $nameNode instanceof Node\QualifiedName && $nameNode->isFullyQualifiedName();
+
+            /** The closest NamespaceDefinition Node */
```
jens1o: why do you use doc comments instead of normal ones?

PR #413, opened 2017-06-16
Remove content

### LINKED ISSUE #397: Don't keep content in memory in addition to AST
@roblourens I see `SourceFileNode` has a `content` property now. But we also store `PhpDocument->content`, which sounds like duplicate memory usage. Can we just remove the `content` property and always rely on the `SourceFileNode`'s `content` property? Or can there be a case where parsing fails so hard the `SourceFileNode` is `null`?

felixfbecker:
Closes #397 

@roblourens

### DIFF
--- a/src/PhpDocument.php
+++ b/src/PhpDocument.php
@@ -47,13 +47,6 @@ class PhpDocument
     private $uri;
 
     /**
-     * The content of the document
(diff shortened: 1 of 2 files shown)


roblourens:
LGTM

PR #415, opened 2017-06-17
Bump tolerant-php-parser

roblourens:
Shouldn't ^0.0.2 install 0.0.3?

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -38,7 +38,7 @@
         "sabre/uri": "^2.0",
         "jetbrains/phpstorm-stubs": "dev-master",
         "composer/composer": "^1.3",
-        "Microsoft/tolerant-php-parser": "^0.0.2"
+        "Microsoft/tolerant-php-parser": "^0.0.3"
     },
     "minimum-stability": "dev",
     "prefer-stable": true,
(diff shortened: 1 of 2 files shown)


jrobeson:
@roblourens : not for 0.0.x versions, only 0.x.y

felixfbecker:
@jrobeson are you sure? At least npm will install 0.0.3 for ^0.0.2 (but not 0.1.0)

felixfbecker:
build is failing, there seems to be an incorrect assertion

jrobeson:
pretty sure.

felixfbecker:
confirmed

jrobeson:
My assertion is actually wrong, as it applies just as well for other 0.x.y releases too. It's when it gets to 1.x.y that it starts acting normally.


> The ^ operator behaves very similarly but it sticks closer to semantic versioning, and will always allow non-breaking updates. For example ^1.2.3 is equivalent to >=1.2.3 <2.0.0 as none of the releases until 2.0 should break backwards compatibility. For pre-1.0 versions it also acts with safety in mind and treats ^0.3 as >=0.3.0 <0.4.0.

PR #416, opened 2017-06-17
WIP support foreach

jens1o:
my first and only approach which does not work in the wild...

### DIFF
--- a/src/DefinitionResolver.php
+++ b/src/DefinitionResolver.php
@@ -543,8 +543,8 @@ class DefinitionResolver
                 ) {
                     return $n;
                 }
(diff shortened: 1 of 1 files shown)


fixtures/foreach.php
```diff
@@ -0,0 +1,4 @@
+<?php
+foreach([0, 1, 2, 3, 4, 5] as $testValue) {
+    $testValue;
```
jens1o: I wanted it to hover the value in the test, but it does not work in the wild.

felixfbecker:
I assume this is superseded by #529

PR #419, opened 2017-06-17
added this completion with tests

vakata:


roblourens:
lgtm

src/DefinitionResolver.php
```diff
@@ -575,7 +575,7 @@ public function resolveExpressionNodeToType($expr)
         //   $myVariable -> type of corresponding assignment expression
         if ($expr instanceof Node\Expression\Variable || $expr instanceof Node\UseVariableName) {
             if ($expr->getName() === 'this') {
-                return new Types\This;
+                return new Types\Object_(new Fqsen('\\' . $this->getContainingClassFqn($expr)));
```
felixfbecker: If you resolve the `this` type statically, how do you want to handle this case?
```php
class Parent {
  /** @return $this */
  public function foo() {
    return $this;
  }
}

class Child extends Parent {
  public function bar() {
    $this->foo()
         ->q|    
  }
  public function qux() {

  }
}

vakata: As previously discussed - this does not support fluent interfaces: https://github.com/felixfbecker/php-language-server/pull/247

From my understanding the idea was to create at least basic ```$this``` autocompletion support and then tackle the more complex cases.

Which leads me to my next question - is it currently even possible to detect the ```$this``` return type? When I tried to implement fluent interfaces I kept getting ```mixed``` (the doc block had no effect).

felixfbecker: Docblocks are parsed by phpDocumentor, which (as you can see in the deleted line) has support for a `$this` return type

felixfbecker: So we could merge this in, but could you add a TODO comment?

PR #420, opened 2017-06-19
Add vendor/validation folders to search.exclude

roblourens:
Also remove `.vscode/` from gitignore. I think it's best practice to share workspace-specific settings in git. Personal preferences can live in user settings.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,7 +1,6 @@
 // Place your settings in this file to overwrite default and user settings.
 {
     "search.exclude": {
-        "**/vendor": true,
         "**/validation": true,
         "**/tests/Validation/cases": true
     }
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,7 +1,7 @@
 // Place your settings in this file to overwrite default and user settings.
 {
-    "search.exclude": {
-        "**/validation": true,
-        "**/tests/Validation/cases": true
-    }
+  "search.exclude": {
+    "**/validation": true,
+    "**/tests/Validation/cases": true
+  }
 }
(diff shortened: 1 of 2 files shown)


.vscode/settings.json
```diff
@@ -0,0 +1,8 @@
+// Place your settings in this file to overwrite default and user settings.
+{
```
felixfbecker: this should follow the .editorconfig

.vscode/settings.json
```diff
@@ -0,0 +1,8 @@
...
+    "search.exclude": {
+        "**/vendor": true,
```
felixfbecker: I am missing `.git/**` inside here

roblourens: Not necessary, because it's included in the default `files.exclude`.

.vscode/settings.json
roblourens: I didn't notice you were using editorconfig. Do you want to have those settings in two places, or just use this for settings that aren't supported by editorconfig?

felixfbecker: Don't care, but the JSON here should have 2 spaces indentation

.vscode/settings.json
felixfbecker: Not neccassarily - I personally have configured vscode to show `.git` because I want to edit files like `.git/info/exclude`

roblourens: Then you should add it to your user `search.exclude`. This doesn't change how that works, right?

felixfbecker: you are right, I thought that this would override it.
I would like to keep `vendor` though, I often search for Node files in quick open to quickly see the properties of a particular Node class.

PR #421, opened 2017-06-19
Fluent interfaces support

vakata:
This is a basic implementation but it seems to work - really looking forward to a review, as I am not sure this is the correct way to go. 
**I have not measured the performance impact of the change!**

### DIFF
--- a/fixtures/completion/this_return_value.php
+++ b/fixtures/completion/this_return_value.php
@@ -1,9 +1,13 @@
 <?php
-class Parent1 {
+class Grand
+{
   /** @return $this */
   public function foo() {
     return $this;
   }
+}
+class Parent1 extends Grand {
+  
 }
 
 class Child extends Parent1 {
--- a/fixtures/completion/this_return_value.php
+++ b/fixtures/completion/this_return_value.php
@@ -1,0 +1,16 @@
+<?php
+class Parent1 {
+  /** @return $this */
+  public function foo() {
+    return $this;
+  }
+}
+
+class Child extends Parent1 {
+  public function bar() {
+    $this->foo()->q
+  }
+  public function qux() {
+
+  }
+}
--- a/fixtures/completion/this_return_value.php
+++ b/fixtures/completion/this_return_value.php
@@ -1,20 +1,23 @@
 <?php
 class Grand
 {
-  /** @return $this */
-  public function foo() {
-    return $this;
-  }
+    /** @return $this */
+    public function foo()
+    {
+        return $this;
+    }
 }
-class Parent1 extends Grand {
-  
+class Parent1 extends Grand
+{
 }
 
-class Child extends Parent1 {
-  public function bar() {
-    $this->foo()->q
-  }
-  public function qux() {
-
-  }
+class Child extends Parent1
+{
+    public function bar()
+    {
+        $this->foo()->q
+    }
+    public function qux()
+    {
+    }
 }
--- a/src/CompletionProvider.php
+++ b/src/CompletionProvider.php
@@ -372,7 +372,7 @@ class CompletionProvider
     }
 
     /**
-     * Adds the FQNs of all parent classes to an array of FQNs of classes
+     * Yields FQNs from an array along with the FQNs of all parent classes
      *
      * @param string[] $fqns
      * @return Generator
--- a/src/CompletionProvider.php
+++ b/src/CompletionProvider.php
@@ -14,6 +14,7 @@ use LanguageServer\Protocol\{
 };
 use Microsoft\PhpParser;
 use Microsoft\PhpParser\Node;
+use Generator;
 
 class CompletionProvider
 {
@@ -374,9 +375,9 @@ class CompletionProvider
      * Adds the FQNs of all parent classes to an array of FQNs of classes
      *
      * @param string[] $fqns
-     * @return string[]
+     * @return Generator
      */
-    private function expandParentFqns(array $fqns)
+    private function expandParentFqns(array $fqns) : Generator
     {
         foreach ($fqns as $fqn) {
             yield $fqn;
--- a/src/CompletionProvider.php
+++ b/src/CompletionProvider.php
@@ -196,18 +196,15 @@ class CompletionProvider
             //    $a->|
 
             // Multiple prefixes for all possible types
-            $prefixes = FqnUtilities\getFqnsFromType(
+            $fqns = FqnUtilities\getFqnsFromType(
                 $this->definitionResolver->resolveExpressionNodeToType($node->dereferencableExpression)
             );
 
-            // Include parent classes
-            $prefixes = iterator_to_array($this->expandParentFqns($prefixes), false);
-
-            // Add the object access operator to only get members
-            foreach ($prefixes as &$prefix) {
-                $prefix .= '->';
+            // Add the object access operator to only get members of all parents
+            $prefixes = [];
+            foreach ($this->expandParentFqns($fqns) as $prefix) {
+                $prefixes[] = $prefix . '->';
             }
-            unset($prefix);
 
             // Collect all definitions that match any of the prefixes
             foreach ($this->index->getDefinitions() as $fqn => $def) {
@@ -232,18 +229,15 @@ class CompletionProvider
             //     TODO: $a::|
 
             // Resolve all possible types to FQNs
-            $prefixes = FqnUtilities\getFqnsFromType(
+            $fqns = FqnUtilities\getFqnsFromType(
                 $classType = $this->definitionResolver->resolveExpressionNodeToType($scoped->scopeResolutionQualifier)
             );
 
-            // Add parent classes
-            $prefixes = iterator_to_array($this->expandParentFqns($prefixes), false);
-
-            // Append :: operator to only get static members
-            foreach ($prefixes as &$prefix) {
-                $prefix .= '::';
+            // Append :: operator to only get static members of all parents
+            $prefixes = [];
+            foreach ($this->expandParentFqns($fqns) as $prefix) {
+                $prefixes[] = $prefix . '::';
             }
-            unset($prefix);
 
             // Collect all definitions that match any of the prefixes
             foreach ($this->index->getDefinitions() as $fqn => $def) {
@@ -388,7 +382,9 @@ class CompletionProvider
             yield $fqn;
             $def = $this->index->getDefinition($fqn);
             if ($def !== null) {
-                yield from $def->getAncestorFQNs($this->index);
+                foreach ($def->getAncestorDefinitions($this->index) as $name => $def) {
+                    yield $name;
+                }
             }
         }
     }
--- a/src/CompletionProvider.php
+++ b/src/CompletionProvider.php
@@ -201,7 +201,7 @@ class CompletionProvider
             );
 
             // Include parent classes
-            $prefixes = $this->expandParentFqns($prefixes);
+            $prefixes = iterator_to_array($this->expandParentFqns($prefixes), false);
 
             // Add the object access operator to only get members
             foreach ($prefixes as &$prefix) {
@@ -237,7 +237,7 @@ class CompletionProvider
             );
 
             // Add parent classes
-            $prefixes = $this->expandParentFqns($prefixes);
+            $prefixes = iterator_to_array($this->expandParentFqns($prefixes), false);
 
             // Append :: operator to only get static members
             foreach ($prefixes as &$prefix) {
@@ -382,18 +382,15 @@ class CompletionProvider
      * @param string[] $fqns
      * @return string[]
      */
-    private function expandParentFqns(array $fqns): array
+    private function expandParentFqns(array $fqns)
     {
-        $expanded = $fqns;
         foreach ($fqns as $fqn) {
+            yield $fqn;
             $def = $this->index->getDefinition($fqn);
-            if ($def) {
-                foreach ($this->expandParentFqns($def->extends ?? []) as $parent) {
-                    $expanded[] = $parent;
-                }
+            if ($def !== null) {
+                yield from $def->getAncestorFQNs($this->index);
             }
         }
-        return $expanded;
     }
 
     /**
--- a/src/Definition.php
+++ b/src/Definition.php
@@ -3,6 +3,7 @@ declare(strict_types = 1);
 
 namespace LanguageServer;
 
+use LanguageServer\Index\ReadableIndex;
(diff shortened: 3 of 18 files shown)


src/DefinitionResolver.php
```diff
@@ -667,13 +666,27 @@ public function resolveExpressionNodeToType($expr)
...
                 if ($expr->parent instanceof Node\Expression\CallExpression) {
-                    $fqn .= '()';
+                    $add .= '()';
                 }
+                $fqn = $classFqn . $add;
                 $def = $this->index->getDefinition($fqn);
                 if ($def !== null) {
                     return $def->type;
+                } else {
+                    $classDef = $this->index->getDefinition($classFqn);
```
felixfbecker: This only works if the implementing class is a direct parent. The parents of the class must be iterated in a loop.

vakata: Sorry if I did not understand correctly - but the loop is right below this line?

felixfbecker: yes, and it iterates the direct parents of `$classDef`. But where does it look at the grandparents?

vakata: > Is there a way to mark the initial type properly as Types\This, and then resolve the type at the point where it is accessed?
I can not think of a way to do this, as sometimes the type is returned right away - without further processing. So if I am in the scope of ```CompletionProvider``` and get a ```Type\This``` node there is nothing I can do with it at that point. But at the same time - resolving a node to type may cause a lot of recursion - in which case ```Type\This``` may be useful until the value is returned to the consumer. To be honest, provided there is no performance hit, my approach would be to convert the type to something concrete as soon as possible (as opposed to dealing with a meta type like ```This```).

vakata: > yes, and it iterates the direct parents of $classDef. But where does it look at the grandparents?

Oh, I don't know why I assumed the ```extends``` key would be expanded, I should have checked - sorry - I will make the necessary adjustments and will definitely check for a performance hit, as there could definitely be one.

src/DefinitionResolver.php
```diff
@@ -670,23 +670,23 @@ public function resolveExpressionNodeToType($expr)
...
-                    if ($classDef !== null && is_array($classDef->extends)) {
-                        foreach ($classDef->extends as $parent) {
-                            $def = $this->index->getDefinition($parent . $add);
-                            if ($def !== null) {
-                                if ($def->type instanceof Types\This) {
-                                    return new Types\Object_(new Fqsen('\\' . $classFqn));
-                                }
-                                return $def->type;
-                            }
+                $lookupDefinition = function (string $base, string $add) use (&$lookupDefinition) {
```
jens1o: Can this be replaced with a method instead? Closures are really slow in php. :/

src/DefinitionResolver.php
```diff
@@ -667,12 +666,26 @@ public function resolveExpressionNodeToType($expr)
...
                 }
-                $fqn = $classFqn . '->' . $expr->memberName->getText($expr->getFileContents());
+                $add = '->' . $expr->memberName->getText($expr->getFileContents());
                 if ($expr->parent instanceof Node\Expression\CallExpression) {
-                    $fqn .= '()';
+                    $add .= '()';
                 }
-                $def = $this->index->getDefinition($fqn);
-                if ($def !== null) {
+                $lookupDefinition = function (string $base, string $add) use (&$lookupDefinition) {
```
felixfbecker: this is essentially `CompletionProvider::expandParentFqns()`, I would refactor that

src/DefinitionResolver.php
vakata: Yes, it should be trivial to move the generator to a method. I just wanted to get an OK on the approach first.

src/DefinitionResolver.php
vakata: Not really - using a generator and looking for the exact method definition guarantees that the execution will break as soon as a match is found. 
Using ```CompletionProvider::expandParentFqns()``` means looking up the whole class hierarchy, storing all definitions in an array, then iterating over it to find matches for the method signature (more definition lookups).

Of course one might consider my approach as premature optimization?

felixfbecker: `CompletionProvider::expandParentFqns()` could be replaced with a generator as its caller only iterates over it. But generators may actually be slower than using an array.

vakata: I guess it depends on the amount of data that needs to be iterated over. But my original point is that here we are looking for a single match, so that is no need to collect all parents and grandparents. 

Of course I am willing to remove the generator, if you think the gains will be negligible - just let me know how to proceed with refactoring since the method in question is in another class? 
(btw, I only included the generator because in the last attempt at fluent interfaces (a few months ago) collecting the data using ```CompletionProvider::expandParentFqns()``` was a bit slow and caused high CPU usage).

felixfbecker: It's fine to keep, but I want to avoid code duplication

felixfbecker:
I would be happy to merge this as soon as the code duplication with CompletionProvider is refactored :)

vakata:
As we discussed in the comments - I do not consider this a duplication - the code pieces are very different:
1) one is a generator, the other one returns an array
2) one is looking for a single match of a method signature - the other for an array of class definitions

If you still believe code from ```CompletionProvider``` should be used I will refactor. As mentioned in the comments above - I will need some pointers as to how to achieve this:
> ... just let me know how to proceed with refactoring since the method in question is in another class?

I would also like to remind you of the other potential problem (again from the comments above):
> ... collecting the data using CompletionProvider::expandParentFqns() was a bit slow and caused high CPU usage

Please, advise on how to proceed?

vakata:
Thank you for the pointers, I will refactor ASAP. I could add a generator to Definition, and materialize the generator in CompletionProvider so that method signatures remain the same.

felixfbecker:
I don't think you need to materialize it (alas `iterator_to_array`). If you look at https://sourcegraph.com/github.com/vakata/php-language-server@53534aff2d647b54f8ec79fbb281e92167af49b2/-/blob/src/CompletionProvider.php#L385:5$references, the result only is iterated over with `foreach`.

vakata:
I will see what I can do - unfortunately I do not think the method is a good fit for ```Definition``` - both ```CompletionProvider``` and ```DefinitionResolver``` use an index in order to expand the parents, which ```Definition``` has no access to - the index could be passed in as a parameter, but I am not sure it is OK to add so much complexity to what currently is a simple value object. I will make the changes and leave the review up to you.

vakata:
Please review the changes (I had to use ```iterator_to_array``` as the results were used by reference). Could you tell me how to test if there is a negative performance impact - are there any performance tests?

src/CompletionProvider.php
```diff
@@ -237,7 +237,7 @@ public function provideCompletion(PhpDocument $doc, Position $pos): CompletionLi
             );
 
             // Add parent classes
-            $prefixes = $this->expandParentFqns($prefixes);
+            $prefixes = iterator_to_array($this->expandParentFqns($prefixes), false);
```
felixfbecker: why is `$def->getAncestorFQNs()` not used directly here?

felixfbecker: The `iterator_to_array()` and `foreach ($prefixes)` could be reduced to a single iteration

src/Definition.php
```diff
@@ -95,4 +96,42 @@ class Definition
...
+            yield $this;
+        }
+        if (is_array($this->extends)) {
+            // iterating once, storing the references and iterating again
+            // guarantees that closest definitions are returned first
+            $definitions = [];
+            foreach ($this->extends as $fqn) {
+                $def = $index->getDefinition($fqn);
+                if ($def !== null) {
+                    yield $def;
```
felixfbecker: I think it would make sense to yield the FQN as a key here so a second function is not needed

src/Definition.php
```diff
@@ -95,4 +96,42 @@ class Definition
...
+                    $definitions[] = $def;
+                }
+            }
+            foreach ($definitions as $def) {
+                yield from $def->getAncestorDefinitions($index);
```
felixfbecker: shouldn't this pass `true`? Otherwise it will be a leaves-only traversal

src/DefinitionResolver.php
```diff
@@ -667,13 +666,21 @@ public function resolveExpressionNodeToType($expr)
...
                 if ($expr->parent instanceof Node\Expression\CallExpression) {
-                    $fqn .= '()';
+                    $add .= '()';
                 }
-                $def = $this->index->getDefinition($fqn);
-                if ($def !== null) {
-                    return $def->type;
+                $classDef = $this->index->getDefinition($classFqn);
+                if ($classDef !== null) {
+                    foreach ($classDef->getAncestorFQNs($this->index, true) as $fqn) {
```
felixfbecker: `getAncestorDefinitions` looks more suitable if you call `getDefinition()` anyway

src/CompletionProvider.php
vakata: It is not used directly as ```$prefixes``` can be an array, as for removing the ```iterator_to_array``` and collecting the modified keys in an array - will do ASAP.

src/Definition.php
vakata: Great idea! Will do.

src/Definition.php
vakata: It is already yielded a few lines above - there is a comment explaining why.

src/DefinitionResolver.php
vakata: I am not sure I get what you mean - here we need the FQN, so that we can append the property / method part onto it and then look for the definition?

vakata:
Please, do not forget about perf testing - how can I test the impact of these changes?

felixfbecker:
You can run https://github.com/felixfbecker/php-language-server/blob/master/Performance.php

roblourens:
Yes, run `git submodule update --init --recursive` then run the Performance.php script with and without your changes. Let me know how it goes and whether you get any useful results from it.

vakata:
I ran the tests on master and on this branch. Here are the results:
```
MASTER:
drupal                   100.00% -  37.64
wordpress                100.00% -   2.72
php-language-server      100.00% -   0.59
tolerant-php-parser      100.00% -  81.62
math-php                 100.00% -   1.06
symfony                  100.00% -  18.05

PR:
drupal                   100.00% -  39.15
wordpress                100.00% -   2.46
php-language-server      100.00% -   0.54
tolerant-php-parser      100.00% -  80.84
math-php                 100.00% -   1.07
symfony                  100.00% -  18.14
```

I see no impact - there are minor differences, but they move up and down with each run. I will use this branch in my VS Code tomorrow to get a feel of the real world usage.

src/Definition.php
vakata: @felixfbecker, do you still need me to do anything here? All the values are yielded in the proper order.

vakata:
Just a quick update - I did not notice any change in responsiveness while using VS Code with the source from this branch, I also did not notice any change in resource usage.

fixtures/completion/this_return_value.php
```diff
@@ -0,0 +1,20 @@
...
+    return $this;
+  }
+}
+class Parent1 extends Grand {
+  
+}
+
+class Child extends Parent1 {
+  public function bar() {
+    $this->foo()->q
```
felixfbecker: could you format these consistently with the rest of the codebase? (4 spaces, brace on new line)

src/Definition.php
felixfbecker: no, looks good

src/CompletionProvider.php
```diff
@@ -382,18 +376,17 @@ public function provideCompletion(PhpDocument $doc, Position $pos): CompletionLi
      * @param string[] $fqns
      * @return string[]
      */
-    private function expandParentFqns(array $fqns): array
+    private function expandParentFqns(array $fqns)
```
vakata: Because it takes an array, not a single definition. There would be identical code looping through the array in both places where this method is used if it is removed. Let me know if you want me to remove it.

src/DefinitionResolver.php
felixfbecker: makes sense

src/CompletionProvider.php
felixfbecker: makes sense.

src/Definition.php
```diff
@@ -95,4 +96,31 @@ class Definition
      * @var string
      */
     public $documentation;
+
+    /**
+     * Gets the definitons of all ancestor classes
+     *
+     * @return Definition[]
```
felixfbecker: Wrong return type, should be `\Generator` (also as a return type hint)

src/Definition.php
```diff
@@ -95,4 +96,31 @@ class Definition
...
+     */
+    public function getAncestorDefinitions(ReadableIndex $index, bool $includeSelf = false)
+    {
+        if ($includeSelf) {
+            yield $this->fqn => $this;
+        }
+        if (is_array($this->extends)) {
```
felixfbecker: you just need to check if it's not `null`

src/Definition.php
```diff
@@ -95,4 +96,31 @@ class Definition
      * @var string
      */
     public $documentation;
+
+    /**
+     * Gets the definitons of all ancestor classes
+     *
```
felixfbecker: `@param` tags are missing

src/CompletionProvider.php
```diff
@@ -382,18 +376,17 @@ public function provideCompletion(PhpDocument $doc, Position $pos): CompletionLi
      * @param string[] $fqns
      * @return string[]
```
felixfbecker: wrong return type, should be `\Generator` (and return type hint too)

src/Definition.php
vakata: I was wondering what to use here - ```\Generator``` or ```\Iterator```, ideally it should be something indicating both the type of the yielded values, like ```\Iterator[Definition]``` but I do not think this is supported. I will add ```\Generator``` for now.

felixfbecker: phpDocumentor has no support for generic types, but you can describe it after the type

felixfbecker:
Thanks a lot!

ADmad:
@vakata Just curious, the benchmarks you showed don't list "Codeignitor" and "cakephp" which are present in the frameworks list [here](https://github.com/felixfbecker/php-language-server/blob/master/Performance.php#L17). Why is that?

vakata:
@ADmad odd, I probably missed something when copying, I will have a look - it was not intentional.

ADmad:
@vakata Thanks, as a CakePHP user I just want to ensure that tests/benchmarks are run for it too :smile:

PR #426, opened 2017-06-21
Restrict workspace/symbol results to non-dependency symbols

felixfbecker:
This improves performance a lot and matches what other language servers do

### DIFF
--- a/src/Server/Workspace.php
+++ b/src/Server/Workspace.php
@@ -33,7 +33,7 @@ class Workspace
      *
      * @var ProjectIndex
      */
-    private $index;
+    private $projectIndex;
 
     /**
      * @var DependenciesIndex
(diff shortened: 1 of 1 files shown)


jens1o:
Could you explain how you improve performance with a simple variable rename and new docs, please?

felixfbecker:
As the title says, this restricts the search to only non-dependency symbols

jens1o:
Okay, thanks. When will you release the new version?

src/Server/Workspace.php
```diff
@@ -84,11 +84,11 @@ public function symbol(string $query): Promise
     {
         return coroutine(function () use ($query) {
             // Wait until indexing for definitions finished
-            if (!$this->index->isStaticComplete()) {
-                yield waitForEvent($this->index, 'static-complete');
+            if (!$this->sourceIndex->isStaticComplete()) {
```
staabm: @jens1o its not just a rename. in some spots its now using `sourceIndex` in others `projectIndex`

jens1o: Yeah, I looked over this, but after he said it I got it.

PR #429, opened 2017-06-22
Fix textDocument/xdefinition

felixfbecker:


jens1o:
I think that that's the reason why my foreach tests does not give any results... happy to see this merged.

felixfbecker:
Pretty sure it's not, VS Code uses `textDocument/definition` not `xdefinition`

jens1o:
And I thought this will fix my issue. Okay, I'll keep debugging why it doesn't show up.

PR #438, opened 2017-07-14
Signature help

vakata:
Initial idea to start the discussion (still needs work)


vakata:
Currently this PR is not ready for merging - I want to get some opinions on the approach and fix as necessary. 

There is one major problem I noticed so far - the currently active parameter is not properly determined (as soon as something is written it shifts one parameter to the left) - I would like some help with that if someone can spare the time and have a look, as I am not very familiar with tolerant parser's objects.

One thing I definitely noticed - this implementation is way cleaner than the one with PhpParser.

src/DefinitionResolver.php
```diff
@@ -234,6 +235,18 @@ public function createDefinitionFromNode(Node $node, string $fqn = null): Defini
             $def->documentation = $this->getDocumentationFromNode($node);
         }
 
+        $def->parameters = [];
+        if (property_exists($node, 'parameters') && $node->parameters) {
```
felixfbecker: can you do this without reflection? i.e. `instanceof`

src/SignatureHelpProvider.php
```diff
@@ -0,0 +1,98 @@
...
+        ) {
+            if ($node instanceof ArgumentExpression) {
+                $arge = $node;
+            }
+            $node = $node->parent;
+        }
+        if (!($node instanceof ArgumentExpressionList) &&
+            !($node instanceof CallExpression)
+        ) {
+            return new SignatureHelp;
```
felixfbecker: A signature without `label` is not valid according to the docblock

src/SignatureHelpProvider.php
```diff
@@ -0,0 +1,98 @@
...
+            }
+        }
+        $def = $this->definitionResolver->resolveReferenceNodeToDefinition($node->callableExpression);
+        if (!$def) {
+            return new SignatureHelp;
+        }
+        return new SignatureHelp(
+            [
+                new SignatureInformation(
+                    trim(str_replace(['public', 'protected', 'private', 'function', 'static'], '', $def->declarationLine)),
```
felixfbecker: Don't like the string manipulation here, you could have a parameter `$function`. I would rather extend `Definition` with another property or method (if possible, without saving new duplicate data to save RAM)

src/Definition.php
```diff
@@ -97,6 +98,12 @@ class Definition
      * @var string
      */
     public $documentation;
+    /**
```
felixfbecker: add an empty line above this

src/DefinitionResolver.php
```diff
@@ -234,6 +235,18 @@ public function createDefinitionFromNode(Node $node, string $fqn = null): Defini
...
+            foreach ($node->parameters->getElements() as $param) {
+                //var_dump($param); die();
+                $def->parameters[] = new ParameterInformation(
+                    $this->getDeclarationLineFromNode($param),
+                    //$param->getName(), // TODO: rebuild this
```
felixfbecker: what does this mean?

src/Server/TextDocument.php
```diff
@@ -411,4 +416,12 @@ public function xdefinition(TextDocumentIdentifier $textDocument, Position $posi
             return [new SymbolLocationInformation($descriptor, $def->symbolInformation->location)];
         });
     }
+
+    public function signatureHelp(TextDocumentIdentifier $textDocument, Position $position): Promise
```
felixfbecker: Please add a docblock for this method with the description from the protocol spec (makes it easier for newcomers to understand the purpose of the method)

felixfbecker:
Maybe @roblourens can help with the parser specifics, he knows more about it than me. Maybe the parameters array gets a `MissingToken` so the count gets confused?

src/DefinitionResolver.php
vakata: I will need @roblourens help for this - I have no idea what to check for. From what I gather it is a trait.

src/DefinitionResolver.php
vakata: I was old code - I am cleaning it up.

src/SignatureHelpProvider.php
vakata: There is no ```label``` property in the ```SignatureHelp``` object - maybe I am missin something - please explain? This is supposed to be an empty result (no nested ```SignatureHelpInformation``` objects).

vakata:
The count should be OK now, but I will need @roblourens help with the ```property_exists``` issue.

src/DefinitionResolver.php
felixfbecker: @roblourens could you add an interface for `FunctionHeader` or callables so this can be checked for with `instanceof`?

src/SignatureHelpProvider.php
vakata: @felixfbecker please provide further information on how I can help improve this? I am still not sure what the issue is.

vakata:
I am afraid the failing build is not my fault - please advise on how to proceed?
The error is in the ```parse-stubs``` script (while parsing ```...phpstorm-stubs/SplType/SplType.php```):
```
Class 'phpDocumentor\Reflection\Types\Mixed' not found
```

I do not get this error locally.

src/SignatureHelpProvider.php
felixfbecker: Sorry, I confused it with the SignatureInformation class

felixfbecker:
weird

roblourens:
I'm not sure what you mean by "the property_exists issue". Your change from the last commit seems good to me - I think it's better to explicitly check the type with `instanceof` instead of assuming that any node with `parameters` is the type you want.

felixfbecker:
@roblourens it would be nice if checking for a callable, function-like node only required a single `instanceof` check for an interface instead of for every possible concrete implementation. PHPParser had a `FunctionLike` interface for this. Given TolerantParser uses a trait for to introduce the `parameters` property, it would make sense to have an interface that the classes that use the trait implement

vakata:
I've been using this branch for a few days now. Issues I've noticed so far:
 - rarely the parameters can not be counted and there is no active parameter highlighted
 - the position is sometimes off ```array_reverse(|explode(``` gives signature help for ```explode```
 - sometimes when moving the cursor fast there is a minimal delay.

I will look into the second issue - I will create a test and try to fix it. 

It would be great if someone else also tried this branch, just so that we can gather opinions (although subjective) about performance - I am using a Surface Pro 4 (i5, 8GB) and experience no issues, but it may be different for someone else.

roblourens:
I do see some relevant helpers in `ParserHelpers.php`, `isFunctionLike` does the same check.

felixfbecker:
I think an interface would be more useful for static analysis if we ever implement type narrowing through `instanceof` like TS does (#168)

felixfbecker:
@vakata do you feel this is stable enough to release?

src/SignatureHelpProvider.php
```diff
@@ -0,0 +1,111 @@
...
+     * Get the short declaration for a callable (class modifiers, function keyword, etc are stripped)
+     *
+     * @param string $declaration
+     * @return string
+     */
+    protected function getShortDeclaration(string $declaration): string
+    {
+        $parts = explode('(', $declaration, 2);
+        $name = array_reverse(explode(' ', trim($parts[0])))[0];
+        return $name . '(' . $parts[1];
```
felixfbecker: this method looks like it would ignore many edge cases, for example `(` or spaces inside a default string value.
Why attempt to explode a declaration string when we have a parser available? I would add the wanted output to the `Definition` class

vakata: I am absolutely certain that this code does not ignore any edge cases - the first ```(``` in a declaration is always after the function name. The function name is always the last token after all modifiers.

I am not sure how to change the current implementation in a robust way - all I need is to remove the modifiers and leave the function name, but keep all parameters (along with their type and default values in place). Doing this using the parser would require quite a huge block of code (from what I gather) especially since there is no way to pretty print the result. 

Basically I would be reconstructing the already available declaration using string concatenation by inspecting the underlying AST just to remove the public / private / abstract / static / function keywords that precede the function name.

vakata:
@felixfbecker as I mentioned in my previous comment:
1) There is a known issue where the position is not calculated correctly, one could argue that any signature help (even with a few edge cases not handled) is better than no signature help. Personally I think the issue should be resolved before merging this, but it is ultimately your call. I will have some time to work on it next week.
2) Performance-wise: I use this branch on a daily basis and have not experienced any issues with speed, but I would be a lot more comfortable if someone else also tested and shared their experience.

src/SignatureHelpProvider.php
felixfbecker: Okay, maybe I'm was just unable to follow this code.
- you split the declaration into two parts, the part before the `(` (function name and keywords) and the rest
- you trim whitespace from the first part (function name and keywords)
- you split the first part by (function name and keywords) by spaces
- you reverse the whole array
- you take the first (initially last) element (function name)
- you concatenate the function name, `(` and the rest

I am pretty sure `end()` is faster to get the last array element than reversing the whole array and taking the first element

felixfbecker:
I don't work on any PHP project currently so I can't do any real usage testing. I don't think a lot of people will try it out by cloning this branch, you could build a .vsix though. Otherwise I would just release it and let people try it out.

src/SignatureHelpProvider.php
vakata: ```end``` requires a reference, which would mean creating a middle step and allocating a separate variable for the array. I am not very familiar with PHP internals, but I believe this means allocating storage in the heap, instead of the stack, so as far as optimization goes - I am not sure which will be faster and more memory efficient. **Anyway - this really is a micro-optimization**. I will create a few tests and change it if necessary.

vakata:
If you feel the issue I described above is minor enough to publish with it and fix later - I am OK with publishing as I am not sure many people would download the .vsix anyway.

jens1o:
well, if somebody builds me such a file or lets me know how I do that, I'm happy to test it. :)

felixfbecker:
@vakata any update?

felixfbecker:
@vakata are you still working on this?

vakata:
@felixfbecker no, I haven't had the time to do any further work. Actually the only thing that was left was the occasional slip up in position calculation, but I doubt I will have time any time soon to work on this.

jens1o:
Could I do something to help you? @vakata

vakata:
@felixfbecker @jens1o sorry I was not around sooner. Currently the only issue I have with this PR is the occasional position miscalculation, but I doubt I will have time to work on it any time soon. I guess the most helpful thing would be for someone else beside me to test this change and share if all worked fine.

felixfbecker:
Closing in favor of #547

PR #441, opened 2017-07-19
depend on phpcs stable version

tradzero:
now stable version of phpcs is 3.0.2


felixfbecker:
3.0 has a different autoloading mechanism. Are you planning to fix the breaking changes?

tradzero:
@felixfbecker sorry, i just wonder why keep this version, i don't have enough ability to fix it, sorry again

Issue #108, opened 2016-10-23
Feature overlay with PHP Integrator


felixfbecker:
@Gert-dev let's move this discussion here so it doesn't clutter the discussion for completion.

> Hi, @felixfbecker, great to hear that someone else is also looking to develop FOSS indexing and IDE-like functionalities for PHP. It seems we are both developing a similar project but for different editors; I maintain the PHP integrator packages for Atom(.io), which also provide things such as autocompletion and code navigation, much like your package does for Visual Studio Code.
> 
> I've already went fairly far in implementing autocompletion and code navigation in my packages and already have most of the functionality you're looking for in this ticket in place (local variable scanning, built-in class fetching, docblock analysis and inheritance, ...). My work can be found here [1]. (I'm explicitly linking to the development version as it went through some major refactoring.)
> 
> I certainly don't want to keep you from writing your own indexer if that's what you desire, but just wanted to let you know that there is a similar project out there. I've only recently split off my PHP indexer (which is also based on php-parser, so I'm very glad to see you got nikic to improve the lexer's robustness when it encounters syntax errors!) from my `atom-base` package into a separate project and repository, hence the apparant low number of commits and activity.
> 
> [1] https://github.com/php-integrator/core/tree/development

felixfbecker:
I saw that you use SQLite. I am interested in how your index structure looks like? Currently this LS just uses a simple `serialize()` to cache the index on disk.
Also, how does your Atom package communicate with your PHP code? As Atom packages are written in CoffeeScript, and the analysis is done in PHP, I imagine you must have invented some kind of protocol to communicate? Which leads me to my final question: What do you think about using the standardized language server protocol instead - e.g. folding your code into this server? Eclipse Che, VS Code, VS and many more use/will use LSP

Gert-dev:
Hi @felixfbecker 

> I am interested in how your index structure looks like?

It currently consists of a [couple of simple SQL tables](https://github.com/php-integrator/core/tree/development/src/Indexing/Sql). I also leverage serializing a bit for things such as function parameters as I do not need access to this data in a further normalized form (i.e. in separate tables) and to avoid a performance hit when joining them.

> Also interested in how you leveraged the PHP parser for completion

I've thought of letting autocompletion results be determined on the PHP side, but as you've probably already found out, php-parser does not deal particularly well with situations where code is not yet complete, such as a user typing `$this->`. For this reason autocompletion is currently completely handled on the CoffeeScript side.

[...]

Issue #69, opened 2016-10-11
Class-likes and functions need to be case-insensitive


felixfbecker:
FQNs for those nodes should be lowercased

Gert-dev:
I wanted to pitch in about this as I've encountered a request for this in my server a while ago. I noticed that composer is not too fond of supporting this in its autoloader, see also [1] and [2]. Because of this, and that varying class names in code is inconsistent and confusing to other programmers, I've decided not to support it for the moment.

To my amazement, I later learned that PHP itself had a trick up its sleeve: some of the class names mentioned in the documentation are *not* using the same casing as the class that is actually in PHP itself (that you would get if you were to e.g. use Reflection). An example of this is `COM`, which is called `COM` according to the documentation but is actually internally reported as being `com` (*sigh*).

My two cents ;-).

[1] https://github.com/composer/composer/issues/1803
[2] https://github.com/composer/composer/issues/4209

PR #451, opened 2017-08-06
Autocomplete speedup

nicolasmure:
Hello,

This PR speeds up the autocomplete response time for the object access operator (`->`) and the static access operator (`::`).

On large projects (with many definitions indexed), these autocomplete features were long to respond because all the indexes were parsed. I added a new way to represent the indexed definition by their namespace (eg `Symfony\Component\HttpFoundation\Request`). So, when requesting for autocomplete definitions of a symbol, we can use this namespace index directly instead of parsing all the indexes.

The new index key is stored in memory only and is not written in the cache file. It is rebuilt when reading the cache file, so it has no negative impacts on memory consumption.

Some metrics (using a symfony 3.1.10 project, see the first commits to figure out how they were recorded) :

Autocomplete requests on `$request->|` :

before : 
```
85 items found, 97766 checks, memory : 549453824 bytes, 2.2225980758667s
85 items found, 97766 checks, memory : 549453824 bytes, 2.1985130310059s
85 items found, 97766 checks, memory : 549453824 bytes, 2.026654958725s
```
after :
```
85 items found, 123 checks, memory : 549453824 bytes, 0.00016593933105469s
85 items found, 123 checks, memory : 549453824 bytes, 0.00015592575073242s
85 items found, 123 checks, memory : 549453824 bytes, 0.00014114379882812s
```

Autocomplete requests on `Response::H|` : 

before :
```
65 items found, 97766 checks, memory : 549453824 bytes, 2.3356730937958s
65 items found, 97766 checks, memory : 549453824 bytes, 2.0599710941315s
65 items found, 97766 checks, memory : 549453824 bytes, 2.2655839920044s
```

after :
```
65 items found, 127 checks, memory : 549453824 bytes, 0.00036478042602539s
65 items found, 127 checks, memory : 549453824 bytes, 0.00025391578674316s
65 items found, 127 checks, memory : 549453824 bytes, 0.00027108192443848s
```

The time to find the items was reduced from ~2s to ~0.0002s, which is 10,000 times faster :rocket: .
It also reduces a lot the CPU consumption because less checks are performed.

I know that there are some feature requests to avoid some directories indexation, but I think that this PR could be useful too to speed up response times ;)


felixfbecker:
Nice!

I have a hard time believing the memory usage though - exactly the same amount of bytes used, even though there is a new array that duplicates the whole index? I wonder whether we could just always use the namespace tree.

nicolasmure:
I've added an other commit to ceate an index representation of the global definitions (reduces the number of checks).
I think that PHP does doe make copies but store references to the definition objects.
As you can see [here](https://github.com/felixfbecker/php-language-server/pull/451/commits/f1419c1701eac825c84264e49ea2d509c9bce48c#diff-18eb38922fb7235266f15163e2b050edR238), I checked the memory consumption usage with `memory_get_usage(true)`.

However, I have some other memory consumption metrics : 
on the current master branch : 
screenshot from 2017-08-07 00-56-30

on this branch : 
screenshot from 2017-08-07 00-58-05

nicolasmure:
I saw that you [had a discussion](https://github.com/felixfbecker/php-language-server/issues/108) with the maintainer of [php-integrator](https://github.com/php-integrator/atom-base), it could be a good idea to store the indexes in a sqlite DB too instead of in-memory and serialized files (will reduce the RAM consumption IMHO).

robclancy:
I switched to using this in vscode and it is significantly faster using 466MiB compared to 454MiB before. Before it meant this was unusable because autocompletes were over a second.

So I'll keep this branch installed for now so I can actually use the add-on to see if I like it.

felixfbecker:
Could you try always using the tree? I could imagine it is not significantly slower than the map

nicolasmure:
Hello @felixfbecker , I'm not sure to get what you're saying :thinking: 
Do you mean to have an index like this : 
```php
$this->definitions['Symfony\Component\HttpFoundation\Request'] => [
    'Symfony\Component\HttpFoundation\Request->getLocale()' => $definition
]
```
and to get rid of the `namespaceDefinitions` index ? It would result in a BC break because the serialized index cache no longer be the same.

felixfbecker:
I mean to get rid of `$definitions` and always use `namespaceDefinitions`.
Changing the cache format is not a BC break, you can simply increase the cache format version.

nicolasmure:
Oh yeah I see, I'll try it then ;)

nicolasmure:
OK @felixfbecker , I added 3 commits to save memory, here are the consumptions (on my testing project):

| Master | Before these 3 commits | After |
| ------------- | ------------- | ------------- |
| 518Mb  | 529Mb  | 522Mb  |

I also yield indexec definitions to save memory instead of copying them in an array.

roblourens:
Do they have an impact on the time? I remember from looking at these earlier, that they can add a surprising overhead.

nicolasmure:
@roblourens , here are some time records grabbed for the [accessor (`->`) operator](https://github.com/nicolasmure/php-language-server/blob/6ba3d6c0c52cd9177d68a3701ef81159dda13aa9/src/CompletionProvider.php#L246-L254) : 

Before these 3 commmits : 
```
0.00016880035400391s
0.00014901161193848s
0.00014591217041016s
0.00020003318786621s
0.00025105476379395s
0.00034809112548828s
0.00015783309936523s
0.00014805793762207s
0.00018405914306641s
0.00019478797912598s
0.00015902519226074s
```

after : 
```
0.0003349781036377s
0.00020289421081543s
0.00030088424682617s
0.00017595291137695s
0.00033283233642578s
0.00017690658569336s
0.0005340576171875s
0.00024986267089844s
0.0003349781036377s
0.00018405914306641s
0.00019097328186035s
```

felixfbecker:
Generators should actually not decrease memory usage because PHP variables are copy-on-write. But it will reduce time because you save an iteration.

staabm:
Iterator also can safe memory when you no longer need to build whole arrays but work item per item (stream like)

felixfbecker:
That's true if the source is an IO stream, but here the source is an already in-memory array.

nicolasmure:
@staabm yes and it is definitively the case here : we're just looping over it, using each item one by one (no need to know how many items are in the iterator, nor to access to the ones next to the currently used).

@felixfbecker as the index used by the `LanguageServer` is a `AbstractAggregateIndex`, it makes sense IMO to use generators to 'stream' definitions from the aggregated indexes.

src/Index/Index.php
```diff
@@ -216,10 +287,53 @@ public function unserialize($serialized)
...
+        return $fqn;
+    }
+
+    /**
+     * Returns the Definitions that are in the given namespace
+     *
+     * @param string $namespace
+     * @return Definition[]
+     */
+    private function doGetDefinitionsForNamespace(string $namespace): array
```
nicolasmure: To avoid logic duplication because I'm using it twice in this class : 
- https://github.com/nicolasmure/php-language-server/blob/6ba3d6c0c52cd9177d68a3701ef81159dda13aa9/src/Index/Index.php#L130
- https://github.com/nicolasmure/php-language-server/blob/6ba3d6c0c52cd9177d68a3701ef81159dda13aa9/src/Index/Index.php#L145

Moreover, this private function returns an array, in opposition to the public function `getDefinitionsForNamespace` which returns a generator. This array is then used in other public functions of this class to yield definitions.

robclancy:
Master: 454MiB
Before the memory saving commits here: 466MiB
Now: 392 MiB

nicolasmure:
@robclancy yes, make sure that the cache has finished to be populated too before making any mesurements. The last commits have changed the cache version so it has the server has to rebuild all the cache items, so memory consumption will start from 0 to x until it has completed the project parsing.

You can see when it has finished to built by wathcing at CPU consumption and also by listing the cache directory by last modification date : 
```bash
$ watch -n 0.5 ls -c ~/.phpls/
```

robclancy:
I'm getting my numbers from the output in vscode after the indexing is done.

robclancy:
Master: [Info  - 11:55:37 pm] All 8950 PHP files parsed in 25 seconds. 454 MiB allocated.

So my numbers above are correct for this project.

robclancy:
I now have...

**This branch**: [Info  - 1:02:48 am] All 8950 PHP files parsed in 17 seconds. 460 MiB allocated.

This is more inline with your numbers, I'm not sure what happened before as this has come from the same cache.

src/Index/Index.php
```diff
@@ -15,14 +15,31 @@ class Index implements ReadableIndex, \Serializable
     use EmitterTrait;
 
     /**
-     * An associative array that maps fully qualified symbol names to Definitions
+     * An associative array that maps namespaces to
+     *     an associative array that maps fully qualified symbol names
+     *     to global Definitions, e.g. :
+     *     [
+     *         'Psr\Log\LoggerInterface' => [
```
felixfbecker: the comment says the first level contains namespaces, but this is an interface

src/Index/Index.php
```diff
@@ -15,14 +15,31 @@ class Index implements ReadableIndex, \Serializable
...
+     *             'Psr\Log\LoggerInterface->log()' => $definition,
+     *             'Psr\Log\LoggerInterface->debug()' => $definition,
```
felixfbecker: What about using a full tree? e.g.
```php
[
  'Psr' => [
    '\Log' => [
      '\LoggerInterface' => [
        '->log()' => $definition
      ]
    ]
  ]
]

File: src/Index/Index.php (hunk 1 above)
Comment (felixfbecker) [inline src/Index/Index.php]: Don't fully see the need for this since it's only `$this->namespaceDefinitions[$namespace] ?? []` but if you want to keep it please rename it to something like `getDefinitionArrayForNamespace` - the current name has no semantic difference to the non-`do` version

File: src/Index/Index.php (hunk 2 above)
Comment (nicolasmure) [inline src/Index/Index.php]: Yes you're right it's a bit confusing. What about renaming the `namespaceDefinitions` to `fqnDefinitions` ? It's also confusing because there's already a
`getDefinition(string $fqn, bool $globalFallback = false)`
function where the `$fqn` parameter is `Psr\Log\LoggerInterface->log()`.

File: src/Index/Index.php (hunk 3 above)
Comment (nicolasmure) [inline src/Index/Index.php]: I'm not sure it worth it as we don't need this kind of representation for now. The actual representation is mainly used to get definitions of the resolved class directly (w/o having to parse its namespace).

File: src/Index/Index.php (hunk 2 above)
Comment (nicolasmure) [inline src/Index/Index.php]: Well, the thing is that everywhere in the app we're using `$fqn` and referring to `fully qualified namespace`, but we're using it for all symbols, ie `Psr\Log\LoggerInterface->log()` and also `Psr\Log\LoggerInterface`.

Which term could describe `Psr\Log\LoggerInterface` ? The `left hand side of access operator` ? Quite long for a name 😝 . What's your opinion on this point ?

File: src/CompletionProvider.php [hunk 4]
```diff
@@ -310,26 +314,22 @@ public function provideCompletion(PhpDocument $doc, Position $pos): CompletionLi
             // Suggest global symbols that either
             //  - start with the current namespace + prefix, if the Name node is not fully qualified
             //  - start with just the prefix, if the Name node is fully qualified
-            foreach ($this->index->getDefinitions() as $fqn => $def) {
+            foreach ($this->index->getGlobalDefinitions() as $fqn => $def) {
```
felixfbecker: This is the only place where that method is used and then the FQN is checked for a prefix. Why doesn't it make use of the namespace tree here?

nicolasmure: Because there was previously a loop on all definitions, with a [check to keep only the global ones](https://github.com/felixfbecker/php-language-server/pull/451/files/1fec8082395e7ce47a905e9acd2d2dedfea50527#diff-18eb38922fb7235266f15163e2b050edL319). By using this new function, we can reduce the loop size by having directly the global definitions (and so return a result faster).

src/Index/Index.php
felixfbecker: `fqn` stands for fully qualified _name_, not namespace 😉

nicolasmure:
I've added a commit to correct the terminology (as you [pointed out](https://github.com/felixfbecker/php-language-server/pull/451#discussion_r141475428) @felixfbecker ).
The serialized cache key has changed (`namespaceDefinitions` => `fqnDefinitions`) so be sure to remove the previous cache files generated on this branch before using this lastest commit.

nicolasmure:
Hello @felixfbecker , any updates on this ? :)

jens1o:
Omg, that would be so awesome! Please! 😅

src/Index/Index.php
felixfbecker: What do you mean with "mainly"? Autocompletion for classes needs to get all classes in a namespace

felixfbecker: Even if it's not a full tree, this should duplicate the prefix on the second level. That's redundant data.

src/Index/AbstractAggregateIndex.php
```diff
@@ -99,20 +99,48 @@ public function isStaticComplete(): bool
...
-                $defs[$fqn] = $def;
+            foreach ($index->getDefinitions() as $fqn => $definitions) {
+                yield $fqn => $definition;
+            }
+        }
+    }
+
+    /**
+     * Returns a Generator providing an associative array [string => Definition]
+     * that maps fully qualified symbol names to global Definitions
```
felixfbecker: global meaning non-members?

src/Index/Index.php
```diff
@@ -15,14 +15,31 @@ class Index implements ReadableIndex, \Serializable
     use EmitterTrait;
 
     /**
-     * An associative array that maps fully qualified symbol names to Definitions
+     * An associative array that maps fully qualified names to
+     *     an associative array that maps fully qualified symbol names
+     *     to global Definitions, e.g. :
```
felixfbecker: I think "global" is incorrect here.

src/Index/Index.php
```diff
@@ -15,14 +15,31 @@ class Index implements ReadableIndex, \Serializable
...
+    private $fqnDefinitions = [];
+
+    /**
+     * An associative array that maps fully qualified symbol names
+     * to global Definitions
      *
      * @var Definition[]
      */
-    private $definitions = [];
+    private $globalDefinitions = [];
```
felixfbecker: Why do we need two arrays?

nicolasmure: This is used to find the global definitions quicker than only having a single array of definitions and checking all the items one by one to get the global definitions.
See [here](https://github.com/felixfbecker/php-language-server/pull/451/files#diff-18eb38922fb7235266f15163e2b050edL319).

jens1o: Does that mean it doubles the needed memory?

nicolasmure: Yes, this array contains references to global definitions (which are already in the `$fqnDefinitions` array). So yes, memory grows up a little bit. I can try to have only the non global definitions in the `$fqnDefinitions`array to avoid the references duplicaiton.

src/Index/Index.php
nicolasmure: Hello,
I started to change the index to use a full tree as you suggested.
I still have to use two arrays to avoid definition collitions : 
```php
[
    'Psr' => [
        '\Log' => [
            '\LoggerInterface' => $definition, // the non member definition
        ],
    ],
]
```
```php
[
    'Psr' => [
        '\Log' => [
            '\LoggerInterface' => [ // here it's an array, not a non-member definition
                '->log()' => $definition, // the member definition
            ],
        ],
    ],
]
```

I haven't pushed it yet as the tests aren't passing on my machine.

felixfbecker: That looks great! Instead of two arrays, you can just use the empty string as a key for the LoggerInterface itself

nicolasmure:
So I'm now using the single (and multi level) definition index array, but even if phpunit tests are passing, there is an error with what is proposed as completion items. I have this problem on nvim and vscode, they're both sending the same request to the server so it does not appear to be a client error.

The problem is that the resolved expression (ie the left hand side of the cursor) is not good. For instance, if you want to see what are the completion possibilities for the `$definition` var of [this function](https://github.com/nicolasmure/php-language-server/blob/e162d94e14114601db77d5ae8943b43c1cbf88ac/src/Index/Index.php#L164), you can add a new line at the beginning of this function, such as 
```php
$definition->get
```
but you'll see that the proposed completion items are not the good one (thay don't belong to the $definition object). Actually, in this exemple they belong to the `$this` object.

I think there might be a n issue with the [node resolution](https://github.com/nicolasmure/php-language-server/blob/e162d94e14114601db77d5ae8943b43c1cbf88ac/src/CompletionProvider.php#L130) ([see here](https://github.com/nicolasmure/php-language-server/blob/e162d94e14114601db77d5ae8943b43c1cbf88ac/src/PhpDocument.php#L215), which use the [microsoft/tolerant-php-parser](https://github.com/Microsoft/tolerant-php-parser)).
The strange point is that I don't have this problem on the master branch :/

Any hints on this ?

felixfbecker:
@nicolasmure if you say there is a newline, why can't we strip it? What is different on master? If you push your WIP I can also take a look

src/Index/Index.php
```diff
@@ -216,10 +272,174 @@ public function unserialize($serialized)
...
+    /**
+     * Splits the given FQN into an array, eg :
+     * 'Psr\Log\LoggerInterface->log' will be ['Psr', '\Log', '\LoggerInterface', '->log()']
+     * '\Exception->getMessage()'     will be ['\Exception', '->getMessage()']
+     * 'PHP_VERSION'                  will be ['PHP_VERSION']
+     *
+     * @param string $fqn
+     * @return array
+     */
+    private function splitFqn(string $fqn): array
```
felixfbecker: Wouldn't this be much simpler with
```php
preg_split("/(\\\\|::|->)/", $fqn, 0, PREG_SPLIT_DELIM_CAPTURE);
```
?

felixfbecker:
@nicolasmure I cannot reproduce your issue:

image

felixfbecker:
@nicolasmure I cannot reproduce your issue:

image

felixfbecker:
Other than https://github.com/felixfbecker/php-language-server/pull/451#discussion_r151031720 I think this would be fine to merge

src/Index/Index.php
staabm: usally the string functions are way faster then preg_*
you would want to measure it I guess

nicolasmure:
Hello @felixfbecker ,
Thank you for trying this out. This is indeed the same use case scenario that I deed, but I only have the completion items for the current class : 
screenshot from 2017-11-15 09-46-54

This PR is rebased on master, I use the tolerant-php-parser v0.0.7, I also did a fresh composer install (removed the previous vendors and lockfile).

src/Index/Index.php
```diff
@@ -216,10 +272,174 @@ public function unserialize($serialized)
...
+        // split fqn at backslashes
+        $parts = explode('\\', $fqn);
+
+        // write back the backslach prefix to the first part if it was present
+        if ('' === $parts[0]) {
+            if (count($parts) > 1) {
+                $parts = array_slice($parts, 1);
+            }
+
+            $parts[0] = sprintf('\\%s', $parts[0]);
```
felixfbecker: why not use plain string concatenation? function calls are expensive

src/Index/Index.php
felixfbecker: Hm, I find it hard to believe that all those `array_slice`, `array_pop` calls etc are faster than a single `preg_split`. Maybe you could commit the code you used to generate the benchmark in the description?

felixfbecker:
Would have loved to merge this today but I sometimes don't get any IntelliSense at all:

image

on master: 

image

There must still be a bug in the logic

nicolasmure:
Hello, I still have the issue too.
I created a [PR on my fork](https://github.com/nicolasmure/php-language-server/pull/1) where I pushed the debug code. This code will log all the messages received and sent on between the client and the server, and will also log the type of node detected on which the cursor is.

Here's the output for the wrong `$definition->get|` suggestions : 
```
{"id":31,"method":"textDocument\/completion","params":{"textDocument":{"uri":"file:\/\/\/home\/nicolas\/www\/php-language-server\/src\/Index\/Index.php"},"position":{"line":165,"character":24}},"jsonrpc":"2.0"}
Expr class = Microsoft\PhpParser\Node\Expression\Variable, fullText = ' $this', uri = 'file:///home/nicolas/www/php-language-server/src/Index/Index.php'
Expr name = this
Content-Type: application/vscode-jsonrpc; charset=utf8
Content-Length: 7222

{"result":{"isIncomplete":true,"items":[{"label":"definitions","kind":10,"detail":"array","documentation":"An associative array that maps splitted fully qualified symbol names\nto definitions, eg :\n[\n    'Psr' => [\n        '\\Log' => [\n            '\\LoggerInterface' => [\n                ''        => $def1, \/\/ definition for 'Psr\\Log\\LoggerInterface' which is non-member\n                '->log()' => $def2, \/\/ definition for 'Psr\\Log\\LoggerInterface->log()' which is a member definition\n            ],\n        ],\n    ],\n]","sortText":null,"filterText":null,"insertText":null,"textEdit":null,"additionalTextEdits":null,"command":null,"data":null},{"label":"references","kind":10,"detail":"string[][]","documentation":"An associative array that maps fully qualified symbol names\nto arrays of document URIs that reference the symbol","sortText":null,"filterText":null,"insertText":null,"textEdit":null,"additionalTextEdits":null,"command":null,"data":null},{"label":"complete","kind":10,"detail":"bool","documentation":null,"sortText":null,"filterText":null,"insertText":null,"textEdit":null,"additionalTextEdits":null,"command":null,"data":null},{"label":"staticComplete","kind":10,"detail":"bool","documentation":null,"sortText":null,"filterText":null,"insertText":null,"textEdit":null,"additionalTextEdits":null,"command":null,"data":null},{"label":"setComplete","kind":2,"detail":"void","documentation":"Marks this index as complete","sortText":null,"filterText":null,"insertText":null,"textEdit":null,"additionalTextEdits":null,"command":null,"data":null},{"label":"setStaticComplete","kind":2,"detail":"void","documentation":"Marks this index as complete for static definitions and references","sortText":null,"filterText":null,"insertText":null,"textEdit":null,"additionalTextEdits":null,"command":null,"data":null},{"label":"isComplete","kind":2,"detail":"bool","documentation":"Returns true if this index is complete","sortText":null,"filterText":null,"insertText":null,"textEdit":null,"additionalTextEdits":null,"command":null,"data":null},{"label":"isStaticComplete","kind":2,"detail":"bool","documentation":"Returns true if this index is complete","sortText":null,"filterText":null,"insertText":null,"textEdit":null,"additionalTextEdits":null,"command":null,"data":null},{"label":"getDefinitions","kind":2,"detail":"\\Generator","documentation":"Returns a Generator providing an associative array [string => Definition]\nthat maps fully qualified symbol names to Definitions (global or not)","sortText":null,"filterText":null,"insertText":null,"textEdit":null,"additionalTextEdits":null,"command":null,"data":null},{"label":"getDescendantDefinitionsForFqn","kind":2,"detail":"\\Generator","documentation":"Returns a Generator that yields all the descendant Definitions of a given FQN","sortText":null,"filterText":null,"insertText":null,"textEdit":null,"additionalTextEdits":null,"command":null,"data":null},{"label":"getDefinition","kind":2,"detail":"\\LanguageServer\\Definition|null","documentation":"Returns the Definition object by a specific FQN","sortText":null,"filterText":null,"insertText":null,"textEdit":null,"additionalTextEdits":null,"command":null,"data":null},{"label":"setDefinition","kind":2,"detail":"void","documentation":"Registers a definition","sortText":null,"filterText":null,"insertText":null,"textEdit":null,"additionalTextEdits":null,"command":null,"data":null},{"label":"removeDefinition","kind":2,"detail":"void","documentation":"Unsets the Definition for a specific symbol\nand removes all references pointing to that symbol","sortText":null,"filterText":null,"insertText":null,"textEdit":null,"additionalTextEdits":null,"command":null,"data":null},{"label":"getReferenceUris","kind":2,"detail":"\\Generator","documentation":"Returns a Generator providing all URIs in this index that reference a symbol","sortText":null,"filterText":null,"insertText":null,"textEdit":null,"additionalTextEdits":null,"command":null,"data":null},{"label":"getReferences","kind":2,"detail":"string[][]","documentation":"For test use.\n\nReturns all references, keyed by fqn.","sortText":null,"filterText":null,"insertText":null,"textEdit":null,"additionalTextEdits":null,"command":null,"data":null},{"label":"addReferenceUri","kind":2,"detail":"void","documentation":"Adds a document URI as a referencee of a specific symbol","sortText":null,"filterText":null,"insertText":null,"textEdit":null,"additionalTextEdits":null,"command":null,"data":null},{"label":"removeReferenceUri","kind":2,"detail":"void","documentation":"Removes a document URI as the container for a specific symbol","sortText":null,"filterText":null,"insertText":null,"textEdit":null,"additionalTextEdits":null,"command":null,"data":null},{"label":"unserialize","kind":2,"detail":"void","documentation":null,"sortText":null,"filterText":null,"insertText":null,"textEdit":null,"additionalTextEdits":null,"command":null,"data":null},{"label":"serialize","kind":2,"detail":"string","documentation":null,"sortText":null,"filterText":null,"insertText":null,"textEdit":null,"additionalTextEdits":null,"command":null,"data":null},{"label":"yieldDefinitionsRecursively","kind":2,"detail":"\\Generator","documentation":"Returns a Generator that yields all the Definitions in the given $storage recursively.\n\nThe generator yields key => value pairs, e.g.\n`'Psr\\Log\\LoggerInterface->log()' => $definition`","sortText":null,"filterText":null,"insertText":null,"textEdit":null,"additionalTextEdits":null,"command":null,"data":null},{"label":"splitFqn","kind":2,"detail":"string[]","documentation":"Splits the given FQN into an array, eg :\n- `'Psr\\Log\\LoggerInterface->log'` will be `['Psr', '\\Log', '\\LoggerInterface', '->log()']`\n- `'\\Exception->getMessage()'`     will be `['\\Exception', '->getMessage()']`\n- `'PHP_VERSION'`                  will be `['PHP_VERSION']`","sortText":null,"filterText":null,"insertText":null,"textEdit":null,"additionalTextEdits":null,"command":null,"data":null},{"label":"getIndexValue","kind":2,"detail":"array|\\LanguageServer\\Definition|null","documentation":"Return the values stored in this index under the given $parts array.\n\nIt can be an index node or a Definition if the $parts are precise\nenough. Returns null when nothing is found.","sortText":null,"filterText":null,"insertText":null,"textEdit":null,"additionalTextEdits":null,"command":null,"data":null},{"label":"indexDefinition","kind":2,"detail":"mixed","documentation":"Recursive function that stores the given Definition in the given $storage array represented\nas a tree matching the given $parts.","sortText":null,"filterText":null,"insertText":null,"textEdit":null,"additionalTextEdits":null,"command":null,"data":null},{"label":"removeIndexedDefinition","kind":2,"detail":"mixed","documentation":"Recursive function that removes the definition matching the given $parts from the given\n$storage array. The function also looks up recursively to remove the parents of the\ndefinition which no longer has children to avoid to let empty arrays in the index.","sortText":null,"filterText":null,"insertText":null,"textEdit":null,"additionalTextEdits":null,"command":null,"data":null}]},"id":31,"jsonrpc":"2.0"}
```

You can see that the `$definition` var is detected as being an instance of `LanguageServer\Index\Index` (localed in `src/Index/Index.php`) : 

```
Expr class = Microsoft\PhpParser\Node\Expression\Variable, fullText = ' $this', uri = 'file:///home/nicolas/www/php-language-server/src/Index/Index.php'
Expr name = this
```
and the var is treated as if it s a `$this`. 

The big json response log is the completion suggestions.

nicolasmure:
I noticed that on the master branch, there is the following [discussion](https://github.com/Microsoft/language-server-protocol/blob/master/protocol.md) between the client and the server when I'm typing `$definition->get|` :
- client : `textDocument/didChange` => server
- server : `textDocument/publisgDiagnostics` => client
- client : `textDocument/completion` => server

but on this branch, the 2nd step is missing :/ I'm gonna investigate on this.

nicolasmure:
I fixed the bug in the [`fix definition removal`](https://github.com/felixfbecker/php-language-server/pull/451/commits/97ec127312069ca9a4a1b78aed6128022361129b) commit, the completion's working on this commit.

Then, I noticed that the completion for non member items was still slow (eg for `array_ma|`) : 
```
1.8270461559296s, 102184 checks, 1 items
```
so I tried to [differentiate the index array](https://github.com/felixfbecker/php-language-server/pull/451/commits/48bbbb5d14a3d1fc01fce7e48383f0ba1690c2bb) to store the member and non member definitions separately to be able to find them quicker.
The result is (for `array_ma|`) :
```
0.13462400436401s, 12392 checks, 2 items
```
as you can see it's way faster. However, there's a bug which duplicates the suggestions for these global defs :/  I still have some work to do on it.

felixfbecker:
Correct me if I'm wrong but isn't the only problem here the fact that the traversal method does a recursive iteration? When requesting completion, isn't it that we only always want the direct children of a given FQN? For example
```php
$obj = new SomeNam|   // I want only top-level symbols (classes/namespaces) to get suggested
$obj = new SomeNameSpace\SomeC // I want only the symbols in SomeNameSpace to get suggested
$obj->me| // I want only members of SomeNamespace\SomeClass to get suggested
```

felixfbecker:
Maybe I'm missing something, but why does the traversal function used for completion have to be recursive? For completion, don't we always just want to suggest the immediate children of a node in the symbol tree? I don't think we need two arrays.

nicolasmure:
Yes, I agree. The completion provides the closest items [from the given fqn](https://github.com/nicolasmure/php-language-server/blob/48bbbb5d14a3d1fc01fce7e48383f0ba1690c2bb/src/CompletionProvider.php#L224). 
The index is interrogated by the completion provider with various fqns. As the definitions in the index are stored in a multi level array, this is why I use recursion.
With your exemple, when typing `$obj->me|`, the index is requested to provide all the definitions of the `SomeNamespace\SomeClass`. These definitions are stored like this way in the index :
```php
[
    'SomeNamespace' => [
        '\SomeClass' => [
            '' // defs go here
        ],
    ],
]
```
this is why the recusive function is for. It uses recusion to get as deep as it can in the tree to find the definitions which are in the given fqn.

For the two arrays, it might look overkill, but it aims to speed the completion by requesting more specifically the definitions. If you look to the [completionProvider](https://github.com/felixfbecker/php-language-server/blob/master/src/CompletionProvider.php#L330-L337) on the master branch, you'll see that it requests all the definitions and then makes a check to test whether the def is member or not. This can be optimized because there are a huge amount of checks performed for only a small amount of definitions passing these checks. I store the defs in these 2 arrays to directly retrieve what I'm looking for instead of having to perform checks.
Have a look to the benchs, you'll see that the check amount is reduced and the response time is also way reduced. It's even more blatant on large projects whit many defs.

felixfbecker:
Yes, so the method can simply traverse into `$definitions['SomeNamespace']['\SomeClass']` (which should have a negligible perf impact) and return those direct children (members) without having to check `isMember` anymore (i.e. keeping the same perf improvements to CompletionProvider). That should not be slower than going into your second array, and it will work with Namespaces, not just members. I am reluctant to special-case members over any other level of the symbol tree.

nicolasmure:
In the [`fix definition removal`](https://github.com/felixfbecker/php-language-server/pull/451/commits/97ec127312069ca9a4a1b78aed6128022361129b) commit (ie before the two arrays), the index tree looks like this : 
```php
[
    'SomeNamespace' => [
        '\SomeClass' => [
            '' => $def1, // the `SomeNamespace\SomeClass` class itself
            '->method()' => $def2, // a method of this class
        ],
    ],
]
```
When calling `getDescendantDefinitionsForFqn('SomeNamespace\SomeClass')` it will yield
```php
'SomeNamespace\SomeClass' => $def1
'SomeNamespace\SomeClass->method()' => $def2
```

This is acceptable when we know what to look for (ie when we have some parts of a namespace), but in [that case](https://github.com/felixfbecker/php-language-server/blob/master/src/CompletionProvider.php#L275-L336), there are too many places to look at (current namespace, global functions, global consts), and there's not always a way to get the defs in these places easily / efficiently. This is why I tried to reduce the number of places to look at by having the 2 arrays.

I agree that it's not pleasant to have these 2 arrays, maybe the `completionProvider` requires some changes for this part, and I'm not really comfortable with it for the moment.

I might look insistent on this part but I'd like to keep the perfs improvements for the most cases :)

felixfbecker:
> but in that case, there are too many places to look at (current namespace, global functions, global consts), and there's not always a way to get the defs in these places easily / efficiently.

So in that case, we have to look at the top namespace (first level of the array) and the current namespace (some level x in the array), so just 2 places to look at. That would be bad if it searched recursively, but if we make the method only search one level, it should be equal perf better as the second member array (actually better, because that doesn't help with the top namespace).

nicolasmure:
Yes, that's why the `completionProvider` requires some changes IMO, it handles many places to look at in a single loop, whereas it should handle one place in one dedicated loop, so we could query the index precisely.

felixfbecker:
I might take a stab at this and if it turns out to be too complicated, we can use the two arrays for the meantime.

felixfbecker:
I changed it to only return direct children in my branch https://github.com/felixfbecker/php-language-server/tree/autocomplet-speedup

case | time for completion
------|----------:
master | `0.1210160255432s`
separate arrays (your branch) | `0.0550549030303s`
only direct children (my branch) | `0.0042061805725s`

looks good? :)

nicolasmure:
Yup, if tests pass then :+1:  :)
The next step would be to also work on [the last part](https://github.com/felixfbecker/php-language-server/blob/fa67f84b73fb8b554d9e45dbc3dd1afe7fba6ecf/src/CompletionProvider.php#L271-L277) of the completion provider to query more specific definitions rather that [all of them](https://github.com/felixfbecker/php-language-server/blob/fa67f84b73fb8b554d9e45dbc3dd1afe7fba6ecf/src/CompletionProvider.php#L326).

It would be good to have some methods in the index (as the `getChildDefinitionsForFqn`), which would be more efficient for this case. As said earlier, it would be good to split this case in smaller parts :) .

I haven't tried your branch yet, just took a look at it for the moment. Thanks for the completion bench :+1:

jens1o:
@felixfbecker Did you already worked on your own banch?

nicolasmure:
Hello @felixfbecker :)
Sorry for the late response, I've checked out your branch and it looks good to me :+1: 
Maybe we could try to merge it as it, as the PR is opened since a while. The [completion improvement for global definitions](https://github.com/felixfbecker/php-language-server/pull/451#issuecomment-346613293) could be resolved in a second time though. WDYT?

felixfbecker:
I'm on the fence whether to merge this in this state or not, and I am guilty of procrastinating the decision. You are right, there are some changes that need to happen at the bottom of CompletionProvider to also not O(n) iterate the whole index (note tests are failing on my branch atm). Which shouldn't be that hard, it might even be easy to just rewrite that part from scratch. If that is done, then this PR would be a very clean improvement for both UX and software architecture.

However: In it's current form I feel like this PR adds a lot of complexity, for improving the completion only for members. The CompletionProvider and DefinitionResolver are definitely the most complex parts of the whole server and low complexity is key for bugfreeness and a low barrier for contributions. We kind of went in circles here by starting with an extra array, then going to a tree, then going back to the separate array. My branch was an attempt to base on the state of the PR a few commits ago when it was still a tree, I just didn't have the time to finish it.

By no means I want to discredit your work, I highly respect the work you put into this and the improvement on the UX site is awesome.

Now if anyone wants to bring that pure-tree implementation over the finish line that would of course make the decision easier :)

Curious to hear your thoughts

Merry Christmas!

nicolasmure:
Thank you for your answer, I agree with you, there were a lot of decision changes in this PR.
I tried to look at the bottom of the CompletionProvider yesterday but as you said it might require a rewrite from scratch, so I wondered if you agreed with the current state of the code.
To be honest I'm using this branch for now on my machine, and it works good (completion is fast), so even if I'd like to help with the rewrite of the CompletionProvider, it's less crucial to me at the moment to spend time and energy on it.
Maybe that now that there are more and more text editors that are adopting LSP, there will have some developers that will have time and motivation to solve this problem :) .

Anyway, thank you for your attention on this PR and for sharing your thoughts :+1: , and also for the maintenance work you're doing here ;)

felixfbecker:
To clarify, I only meant rewriting that very last section of CompletionProvider (that provides completion for global names).

robclancy:
Current stable is basically unusable because of how slow it is... I haven't been following the code here but shouldn't the end user experience be more important than how complicated the code is?

Since this isn't directly an add-on but installed by the add-on would it be possible to put this on a beta branch and then in the add-on have a "use beta language server" option?

Declspeck:
I made the requested change in [my branch](https://github.com/Declspeck/php-language-server/tree/autocomplet-speedup).

Turns out I can't push commits to this PR, and I wouldn't really like to open a new PR because that would make it seem like I did all the hard work, not @nicolasmure. 

Benchmarks before and after (`benchmarks/completion.php`, as modified in my branch):

Before:
```
Getting completion
Time ($this->|): 0.0012650489807129s
90 completion items
Time (new|): 0.095192909240723s
1 completion items

```

After:
```
Getting completion
Time ($this->|): 0.0012638568878174s
90 completion items
Time (new|): 0.00032806396484375s
1 completion items
```

Declspeck:
@felixfbecker  - I also started to add support for the case below (no commits yet)

```php
use Microsoft\PhpParser as TheParser;
TheParser\No|
```

However, I feel like the code is currently pretty hard to read. Would you be against adding a class for representing a class or namespace name, and which could provide useful operations such as:

- Implement `\IteratorAggregate`, `\ArrayAccess`, `\Countable`
- `isFullyQualified()`
- `getFirstPart()` and `getLastPart()`
- `getParent()`
- `startsWith()`
- `withAppended($anotherName)`
- `withPrefixReplacedWith($prefixName, $replacementName)`
- `toString()`, `fromString($str)`
- `fromResolvedName(PhpParser\ResolvedName $name, )`
- `fromQualifiedName(PhpParser\Node\QualifiedName $name)`
- `fromToken(PhpParser\Token, string $fileContents)`
- `fromArray($array, $isFullyQualified)`

The `ResolvedName` class from `PhpParser` is not up to this task.

Declspeck:
I just realized one thing - the current Index stores both `namespace` and `class` definitions in the same tree. This causes problems when you have a name which is both a `class` and a `namespace`, e.g. `Microsoft\PhpParser\Node`. The index needs to be modified so that the same name may refer to multiple definitions, otherwise completion at `use \Microsoft\PhpParser\N|` cannot be reliably implemented.

felixfbecker:
@Declspeck For that we just need to make sure FQNs for namespaces cannot clash with classes, which I would do by simply having namespace FQNs end in a backslash

felixfbecker:
Re: QualifiedName class, we already have to interact with a lot of implementations of those from the parser, phpDocumentor etc. so I found just using strings much easier (and probably more performant) so far instead of introducing a third one (and, for example, you don't need `equals()` to compare strings). Could the methods you would wanna add just be pure functions maybe?

Declspeck:
@felixfbecker Pure functions would also work, I'll try that.

As for equals - for pure strings, you'd still need to use `strcasecmp` or something if you'd want to support case-insensitivity of namespace or function names. And if the name is `\Foo\Bar::$c`, you'd need to 
 do a case-insensitive comparison on the leading part and a case-sensitive comparison on the trailing part, so an equals function would still be needed.

Adding a trailing backslash to namespaces seems like a good idea, I'll try that and see where it leads.

felixfbecker:
Whenever you have something you feel like is ready to be merged I would encourage you to open a PR!

For case insensitivity please see https://github.com/felixfbecker/php-language-server/issues/69

hubertprein:
I'd love to see this getting merged into the master branch in the very near future. The current version that's also being used with Visual Studio Code, using the auto-completion, takes to long to come up with suggestions.

Declspeck:
I could not edit this PR, so I opened a new PR: #599

felixfbecker:
Thanks for your fantastic work @Declspeck @nicolasmure, merged in 24388bcf26d6e088b132904ab81f02a4df813212

nicolasmure:
Thank you @felixfbecker and the contributors :D

PR #459, opened 2017-08-14
Fix memory leak issue

john-nguyen09:
Hi,

Just spotted the reason why a lot of people report that there is a memory leak issue (I personally encountered this countless times), it is just a simple fix.

Never created a pull request before, so please let me know if something is needed to be done.

### DIFF
--- a/src/FqnUtilities.php
+++ b/src/FqnUtilities.php
@@ -22,7 +22,7 @@ function getFqnsFromType($type): array
     }
     if ($type instanceof Types\Compound) {
         for ($i = 0; $t = $type->get($i); $i++) {
-            foreach (getFqnsFromType($type) as $fqn) {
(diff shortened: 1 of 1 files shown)


felixfbecker:
Wow, great catch. A test would be nice for this.

roblourens:
Oh man, good catch indeed!

src/FqnUtilities.php
```diff
@@ -22,7 +22,7 @@ function getFqnsFromType($type): array
     }
     if ($type instanceof Types\Compound) {
         for ($i = 0; $t = $type->get($i); $i++) {
```
ADmad: Wouldn't it be better to avoid assignment in the `foreach` statement and use `do .. while` instead? That would make the code more readable.

felixfbecker: Don't understand what you mean

PR #484, opened 2017-09-28
Fix missing diagnostics for nodes

MadHed:
Fixes a bug where nodes that are inside a children array were not checked for diagnostics.
Also adds support for different parser severity levels.

### DIFF
--- a/src/TreeAnalyzer.php
+++ b/src/TreeAnalyzer.php
@@ -52,8 +52,34 @@ class TreeAnalyzer
         $this->collectDefinitionsAndReferences($this->sourceFileNode);
     }
 
+    private function collectDiagnostics($node) {
+        if (($error = PhpParser\DiagnosticsProvider::checkDiagnostics($node)) !== null) {
+            $range = PhpParser\PositionUtilities::getRangeFromPosition($error->start, $error->length, $this->sourceFileNode->fileContents);
+
+            if ($error->kind == \Microsoft\PhpParser\DiagnosticKind::Error) {
+                $severity = DiagnosticSeverity::ERROR;
+            }
+            else {
+                $severity = DiagnosticSeverity::WARNING;
+            }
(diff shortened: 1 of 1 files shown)


src/TreeAnalyzer.php
```diff
@@ -52,8 +52,34 @@ public function __construct(PhpParser\Parser $parser, string $content, DocBlockF
...
+    private function collectDiagnostics($node)
+    {
+        if (($error = PhpParser\DiagnosticsProvider::checkDiagnostics($node)) !== null) {
+            $range = PhpParser\PositionUtilities::getRangeFromPosition($error->start, $error->length, $this->sourceFileNode->fileContents);
+
+            if ($error->kind == \Microsoft\PhpParser\DiagnosticKind::Error) {
+                $severity = DiagnosticSeverity::ERROR;
+            } else {
+                $severity = DiagnosticSeverity::WARNING;
+            }
```
felixfbecker: could you use a `switch` for this? that would it clearer that you are doing a 1:1 mapping

src/TreeAnalyzer.php
```diff
@@ -52,8 +52,34 @@ public function __construct(PhpParser\Parser $parser, string $content, DocBlockF
         $this->collectDefinitionsAndReferences($this->sourceFileNode);
     }
 
+    private function collectDiagnostics($node)
```
felixfbecker: add a docblock

src/TreeAnalyzer.php
MadHed: Sure

MadHed:
TBH, I would actually like to refactor out the mutual recursion between `collectDefinitionsAndReferences` and `update`. What are your thoughts on that?

felixfbecker:
If the resulting code is clearer, sure.

PR #486, opened 2017-10-02
Exclude non-essential files in .gitattributes

vinkla:
[Exclude non-essential files in .gitattributes](https://www.reddit.com/r/PHP/comments/2jzp6k/i_dont_need_your_tests_in_my_production/). This way users of this package doesn't have to download non-essential files such as the `tests` directory or the `.travis.yml` file.

Happy Hacktoberfest!

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -4,8 +4,10 @@
 /fixtures export-ignore
 /tests export-ignore
 /images export-ignore
+/validation export-ignore
 /.gitattributes export-ignore
 /.gitignore export-ignore
+/.gitmodules export-ignore
 /.dockerignore export-ignore
 /.travis.yml export-ignore
 /codecov.yml export-ignore
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,17 @@
+* text=auto
+
+/.vscode export-ignore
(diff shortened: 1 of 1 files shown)


.gitattributes
```diff
@@ -0,0 +1,19 @@
...
+/.gitmodules export-ignore
+/.dockerignore export-ignore
+/.travis.yml export-ignore
+/codecov.yml export-ignore
+/Dockerfile export-ignore
+/Performance.php export-ignore
+/php.ini export-ignore
+/phpcs.xml.dist export-ignore
+/phpunit.xml.dist export-ignore
+/README.md export-ignore
```
felixfbecker: Isn't a README always good to have in a distribution?

vinkla: It is common to exclude this file as well. Think of this as when the package is installed in production. When there is no need to check the `vendor` directory.

felixfbecker: True, but does this setting only apply to production environments?

vinkla: Yes, this will apply if the user installs packages with the `--prefer-dist` flag or config.

felixfbecker: gotcha, that makes sense to me.

PR #487, opened 2017-10-02
Update composer.json structure

vinkla:
I've updated the structure of the `composer.json` file:

- Added config to prefer distribution versions of packages.
- Removed the `type` since it defaults to library.
- Renamed `Microsoft` to `microsoft`.
- Sorted the required packages by name.

Let me know what you think and if I should update anything. 

Happy Hacktoberfest!

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -61,7 +61,6 @@
         "post-install-cmd": "@parse-stubs"
     },
     "config": {
-        "preferred-install": "dist",
         "sort-packages": true
     },
     "minimum-stability": "dev",
--- a/composer.json
+++ b/composer.json
@@ -1,14 +1,7 @@
 {
     "name": "felixfbecker/language-server",
     "description": "PHP Implementation of the Visual Studio Code Language Server Protocol",
-    "authors": [
-        {
-            "name": "Felix Becker",
-            "email": "felix.b@outlook.com"
-        }
-    ],
     "license": "ISC",
-    "type": "library",
     "keywords": [
         "php",
         "language",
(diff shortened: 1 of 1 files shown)


felixfbecker:
> Added config to prefer distribution versions of packages

Why that change? It has the disadvantage of not having the dependency in vendor/ be a git repo that can be hacked on

vinkla:
> Why that change? It has the disadvantage of not having the dependency in vendor/ be a git repo that can be hacked on.

I'm not sure what you're referring to. This update shouldn't have a bad impact on the current behaviour.

felixfbecker:
When I run `composer install` in my dev environment, Composer will by default `git clone` every dependency, allowing me to make edits in my `vendor` directory (e.g. on the JSON RPC library), commit and push them there. With `--prefer-dist`, it will download a tarball from GitHub, which I cannot commit in. This makes the workflow a lot worse since Composer doesn't have something like `npm link`. I don't see the need for that configuration to sit in composer.json, if you are in a production environment you can pass `--prefer-dist` to Composer to get the same effect (we do that in Travis)

felixfbecker:
At least here: https://github.com/felixfbecker/vscode-php-intellisense/blob/master/.travis.yml#L17
We should add it in this repo too

vinkla:
I've pushed a fix to remove the distribution configuration.

felixfbecker:
Could you add the `--prefer-dist` in travis.yml?

vinkla:
> Could you add the --prefer-dist in travis.yml?

Sure, I'm working on a pull request for the `.travis.yml` file right now.

vinkla:
> Could you add the --prefer-dist in travis.yml?

Sure, I'm working on a pull request for the `.travis.yml` file right now. Just waiting on this pull request to be completed first.

PR #488, opened 2017-10-02
Update phpunit config

vinkla:
This pull request updates the PHPUnit XML configuration: 

- Added more strict configuration.
- Added coloured output from our tests.
- Don't allow risky tests and warnings.

Happy Hacktoberfest!


felixfbecker:
Thanks!

vinkla:
@felixfbecker any pointers on how we can fix the risky test?

felixfbecker:
The test is fine, is it possible to add an ignore comment to surpress the error?

vinkla:
> The test is fine, is it possible to add an ignore comment to surpress the error?

I've pushed an fix to ignore risky tests. It should do the job.

PR #502, opened 2017-10-25
Drop `composer run-script parse-stubs` from readme

NateEag:
As the parse-stubs step is done automatically by `composer install` since https://github.com/felixfbecker/php-language-server/commit/34d3d2030dabc29bd919d0ea795bcb8d994228de, we no longer need to explicitly instruct people to do it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -199,3 +199,7 @@ Run the tests with
 Lint with
 
     composer lint
+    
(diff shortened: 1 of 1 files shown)


felixfbecker:
It should still be mentioned that this may need to be run when you make changes to the code

NateEag:
Good point.

That would only be necessary after updating the JetBrains PHPStorm stubs via Composer, right?

Or am I missing something about the process?

felixfbecker:
also when you make any kind of change to the index, parsing etc

NateEag:
Does the updated patch look okay?

PR #505, opened 2017-10-28
use FunctionLike Interface

jens1o:
because why not?

### DIFF
--- a/src/ParserHelpers.php
+++ b/src/ParserHelpers.php
@@ -40,10 +40,7 @@ function getFunctionLikeDeclarationFromParameter(Node\Parameter $node)
 
 function isFunctionLike(Node $node)
 {
-    return
-        $node instanceof Node\Statement\FunctionDeclaration ||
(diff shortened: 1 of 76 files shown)


jens1o:
Why does that decrease coverage, while actually cleaning up and using all features of the PhpParser

MadHed:
@jens1o Because you removed 2 lines, meaning the uncovered lines have a relatively bigger impact. ;)

jens1o:
Ah, that's making sense. Thanks :)

src/ParserHelpers.php
```diff
@@ -40,10 +40,7 @@ function getFunctionLikeDeclarationFromParameter(Node\Parameter $node)
 
 function isFunctionLike(Node $node)
```
felixfbecker: we can get rid of this function then

jens1o: Okay, I though we're leaving it as a shortcut. Will do this tomorrow.

PR #518, opened 2017-11-08
Update README.md

bmax:
Update README.md for instructions on using xdebug


bmax:
@felixfbecker should I go more into detail with using the  [VSCode PHP debug](https://github.com/felixfbecker/vscode-php-debug) plugin?

PR #528, opened 2017-11-15
Static analysis: Using $this in a static method should show an error

### LINKED ISSUE #476: Static analysis: Using $this in a static method should show an error
Hi there,

I've checked this situation in Atom IDE 1.21 beta, with ide-php package installed. It should check for this incorrect usage of $this.

MaartenStaa:
Closes #476 

Two scenarios produce errors, either $this is used in a static method (as described in the issue), or when it's used outside of a class method altogether, e.g.:

```php
<?php

class A
{
    public static function foo()
    {
        return $this; // error here
    }

    public function bar()
    {
        return $this; // no error here (of course)
    }
}

function foo()
{
    return $this; // error here
}

echo $this; // error here
```


felixfbecker:
Could you add a test for this?

jens1o:
Should this be included in the parser instead?

Issue #200, opened 2016-12-04
Register foreach variables


felixfbecker:

felixfbecker:
@jens1o are you working on this?

jens1o:
Yes, but I'm not sure whether I forgot anything, because I couldn't test it out, because I don't know how I could achieve that... :/

felixfbecker:
Why couldn't you test it?

jens1o:
I don't know how. The symlinks does not seem to work unfortunately. This way, I only could write tests, but I don't want to create a pr where I can't test it out in the wild, so something like that `define` things won't happen again where I missed something that only happens in the wild.

felixfbecker:
What do you mean with the symlinks do not work? What symlinks?

jens1o:
Symlinks that references to my clone... How am I supposed to test it out, maybe I did something wrong? But this was my one-and-only approach.

MadHed:
Are you talking about linking your working copy into the .vscode/extensions folder?

jens1o:
yep, but Felix gave me a heads up, so it should work now. But thanks for wanting to help :)

MadHed:
alright

jens1o:
will create the pr(depending whether I need to add something) tomorrow or on Saturday.

NateEag:
@jens1o, any progress on this? I just ran into this issue myself.

jens1o:
Actually, I didn't touched it seriously anymore since the new parser. I'll take a look now.

jens1o:
@felixfbecker https://github.com/jens1o/php-language-server/commit/5c9b155004946f6af3315650354f35651169151d
Where do I need to register it, too? I think it isn't complete, yet.

felixfbecker:
That's it for go to definition, I think you need to implement it in the CompletionProvider too

jens1o:
Unfortunately `Go to Definition` doesn't works either. :/

jens1o:
Okay, it works now.

NateEag:
Nice. Thanks for working on this, @jens1o!

PR #529, opened 2017-11-16
[WIP] register foreach variables

jens1o:
#200


src/DefinitionResolver.php
```diff
@@ -970,6 +979,11 @@ public function getTypeFromNode($node)
             return $this->resolveExpressionNodeToType($node->argumentExpressionList->children[2]->expression);
         }
 
+        // FOREACH KEY VARIABLE
+        if ($node instanceof Node\ForeachKey) {
+            return new Types\Integer;
```
felixfbecker: The key of a `foreach` variable doesn't have to be integer. It can be anything. You need to find out by resolving the iterated variable to a type and if it's an array get its `keyType`

felixfbecker:
Doe this work for this?
```ts
class MyObj {
  public $foo;
}
$obj = new MyObj();
$arr = [$obj];
foreach ($arr as $val) {
  $val->f // get completion and jump to definition for foo
}
```

src/DefinitionResolver.php
staabm: Does this need special treatment for

`foreach($arr as list($a1, $a2))` 

Like looks?

jens1o: Oops, that was just for testing purposes, and indeed, yes, I think `list($a1, $a2)` needs special treatment, although I do not know how I can achieve that.

jens1o:
Not yet, unfortunately. How can I achieve that, too?

Issue #536, opened 2017-11-20
Setting to disable "$this can only be used in an object context."


felixfbecker:
_From @Gauss23 on November 20, 2017 15:23_

Hello,

we just wondered about a lot of "syntax"-errors in our code. The message is:

> $this can only be used in an object context.

Before your update to version 2.0.0 everything was fine. The files IntelliSense is now marking as wrong are config files, which are required/included in object context, so they are perfectly working. Of course IntelliSense cannot know that. Is there a chance to make this check optional?

Best regards,
Christopher

_Copied from original issue: felixfbecker/vscode-php-intellisense#201_

felixfbecker:
@Gauss23 is there a way to identify these config files? E.g. do they have a top-level return?

Gauss23:
Unfortunately no, we are using this in config files and templates. Our renderer is an object, the templates it is parsing are PHP files and are using $this for multiple reasons. PHP is not complaining and everything is working. We don´t want to make too much trouble about it,.

A simple config switch would be sufficient for us. For the majority of the people, the error will be correct.

felixfbecker:
Want to do a PR?

Gauss23:
I just looked into the sources. I can see how to add a config var in extension.ts and how to push it as a cmd option to bin/php-language-server.php.
How do you want to push the config variable to the TreeAnalyzer? I saw it´s created by PhpDocument.php.
Using a global would not be the preferred method I think. Is there already some kind of infrastructure to pass something from the outside deeply into the the objects? 
Maybe there will be more config settings in the future. 

Best regards

felixfbecker:
No, it would have to implement `didChangeConfiguration`: https://github.com/Microsoft/language-server-protocol/blob/master/protocol.md#workspace_didChangeConfiguration

Gauss23:
I thought I can add a line in: `vscode-php-intellisense/src/extension.ts`
Like this line:
`    const memoryLimit = conf.get<string>('memoryLimit') || '-1';`
I would add:
`    const checkThisInClasses = conf.get<string>('checkThisInClasses') || '1';`

And then add a command option in the same file.

Can´t see how to implement didChangeConfiguration, sorry.

felixfbecker:
That wouldn't help. The configuration can change.

Gauss23:
At the moment you´re not handling config changes, at least I can´t find an implementation of this didChangeConfiguration in intelliSense.
I´ll try to find an example for this implementation.

[...]

PR #543, opened 2017-11-30
Tolerate $this usage in anonymous functions

alWerewolf:
Alternative solution for issue #536.
For $this var usage it gives warning
>$this might not be bound by invoker of callable or might be bound to object of any class. Consider adding instance class check.

And hints developer to check class of $this before usage.
Class check can be added in form 
```php
if ($this instanceof [class])" {
    /*use $this*/
}
```
Note: if class check added no type checking is done in language server it just hides the warning.
There is a leaf in code marked by comment:
> //else we should warn here that invoker can bind $this to object of any class. Should we?

Question in comment is for community.


felixfbecker:
I've never seen anyone use `if $this instanceof` so I don't think anyone would actually use that. I don't think it belongs into the language server.

jens1o:
@felixfbecker Here you go: https://github.com/pmmp/PocketMine-MP/blob/1f5f67e08712078588ef2241d870a0712b7a0cf4/src/pocketmine/block/Liquid.php#L182-L190

alWerewolf:
$this can be bound to ANY object by caller of Closure(anonymous function) so we must warn about it.
It's all about developer must know for sure how his function is called and which object is bound.
Class check for $this var must ensure expected results.
Anyone can suggest their way to do class check. I think there are many ways.
I've choosen $this instanceof because of simplicity and it's favor vs deprecated is_a.
I thought of example where you pass your anonymous function to "blackbox".
You never know for sure how it's called so checking $this var will help from getting into trouble.

felixfbecker:
I decided for disabling the warnings in these cases instead. This is similar to what e.g. TypeScript does - `this` is `any` by default, so no checks. Only in cases where it is known to be a certain type (or type hinted) should we do warnings. In the future, we can give the user the option to type hint `$this`, but until then we should force the user to write code to circumvent the warning.

PR #545, opened 2017-12-03
Update checking of $this usage to only error in static methods.

MaartenStaa:
This rolls back a few of the changes in #528 to address the concerns in #536. It will now only produce an error in static methods, and leave usages of $this in the root and in closures alone.


jens1o:
Personally, I'd prefer the approach in https://github.com/felixfbecker/php-language-server/pull/543

PR #550, opened 2017-12-11
Fix return self completion

phil-nelson:
I found that functions that returned `self` would not complete properly. This fixes resolving non-static and static functions that return `self`.


tests/Server/TextDocument/CompletionTest.php
```diff
@@ -716,6 +716,28 @@ public function testMethodReturnType()
         ], true), $items);
     }
 
+    public function testMethodReturnSelf()
```
felixfbecker: I find it a bit weird that these are tests of `textDocument/completion`. Clearly what's being tested here is the DefinitionResolver?

phil-nelson: Makes sense, I've removed these tests and the cases are now covered with the validation cases. Is that what you meant?

felixfbecker: I was thinking of a unit test for DefinitionResolver, but this seems sufficient!

PR #551, opened 2017-12-12
Foreach completion

### LINKED ISSUE #200: Register foreach variables

phil-nelson:
This adds some basic support for completing `foreach` keys/values. For a complete list of what works have a look at the test, but I believe most "normal" cases work.

Limitations:

* Using `list()` for values is unsupported. This isn't straightforward, and as far as I can tell we don't support a straight up `list()` (outside a foreach) for completion
* No support for iterating public object properties
* No support for `\Traversable` objects

I think implementing these limitations is probably more significant work than what's here. I think the cases that work are probably the most used/popular, so even given these limitations I think this would be worth having. We can work on the missing bits in the future.

Closes #200

### DIFF
--- a/src/DefinitionResolver.php
+++ b/src/DefinitionResolver.php
@@ -1097,14 +1097,14 @@ class DefinitionResolver
             return new Types\Mixed_;
         }
 
-        // FOREACH KEY
+        // FOREACH KEY/VARIABLE
         if ($node instanceof Node\ForeachKey || $node->parent instanceof Node\ForeachKey) {
             $foreach = $node->getFirstAncestor(Node\Statement\ForeachStatement::class);
             $collectionType = $this->resolveExpressionNodeToType($foreach->forEachCollectionName);
             if ($collectionType instanceof Types\Array_) {
                 return $collectionType->getKeyType();
             }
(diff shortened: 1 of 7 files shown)


src/DefinitionResolver.php
```diff
@@ -1080,6 +1097,27 @@ public function getTypeFromNode($node)
...
         }
 
+        // FOREACH KEY/VARIABLE
+        if ($node instanceof Node\ForeachKey || $node->parent instanceof Node\ForeachKey) {
+            $foreach = $node->getFirstAncestor(Node\Statement\ForeachStatement::class);
+            $collectionType = $this->resolveExpressionNodeToType($foreach->forEachCollectionName);
+            if ($collectionType instanceof Types\Array_) {
+                return $collectionType->getKeyType();
+            }
+            return new Types\Mixed_();
```
jens1o: code style; no brackets at the end.

andreymrt:
@phil-nelson can you please take a look at #561 could be related to your PR on foreach completion.
Following code snippet makes the lang server throw an exception
```php
<?php
foreach($Calendar as $day)
	echo $day->Date->format("D");
```
While, this one is fine:
```php
<?php
foreach($Calendar as $day)
	echo $day->format("D");
```

PR #567, opened 2017-12-27
Fix infinite loop in foreach resolving

### LINKED ISSUE #566: PHP Intellisense, huge memory leak when building cache
We use vscode and php intellisense with Laravel project and we have a huge memory leak (16 Go in 5 minutes) with this file : **vendor/symfony/http-kernel/DependencyInjection/RegistrerControllerArgumentLocatorsPass.php** during building cache.

If I renamed this file  .OLD , all cache will be built without any problem. If i renamed it .php after the cache is build, all is OK.

My version : 

Version 1.19.0
Commit 816be6780ca8bd0ab80314e11478c48c70d09383
Date 2017-12-14T09:56:48.842Z
Shell 1.7.9
Renderer 58.0.3029.110
Node 7.9.0
Architecture x64
PHP Intellisense 2.2.2
Laravel Project

phil-nelson:
There is a bug that occurs when we come across code like:

```php
foreach ($a->getArray() as $a) {
```

Where in `resolveVariableToNode`, when we try to find where `$a` was defined, we have to traverse the AST up. So if we are within the loop and we come across `$a`, we traverse up and find the `foreach` statement, and find it is defined as the forEachValue. Then we want to find where that was defined, and we get to the forEachCollection (`$a->getArray()`) and we again have to traverse up the tree to find $a. We come come across the `foreach` statement again - and there isn't a way to identify that we got there from within the `foreach` statement, so we do the same as if we came from within the foreach loop and check the key/value, find it in the value... and we're stuck in a loop.

The only way I've found to fix this is identify if we get to the `foreach` statement from the the `forEachCollectionName`, in which case we do not want to check the key/value.

Closes #566 
Closes felixfbecker/vscode-php-intellisense#192

### DIFF
--- a/fixtures/completion/foreach.php
+++ b/fixtures/completion/foreach.php
@@ -38,3 +38,10 @@ foreach ($bar->test() as $value) {
 
 foreach ($unknownArray as $unknown) {
     $unkno
+}
+
+foreach ($loop as $loop) {
+}
+
+foreach ($loop->getArray() as $loop) {
+}
--- a/src/DefinitionResolver.php
+++ b/src/DefinitionResolver.php
@@ -571,10 +571,27 @@ class DefinitionResolver
 
             // If we get to a ForeachStatement, check the keys and values
             if ($n instanceof Node\Statement\ForeachStatement) {
-                if ($n->foreachKey && $n->foreachKey->expression->getName() === $name) {
+                // Only check keys and values if we did not get here from the foreach collection, otherwise code like
+                // foreach ($a as $a) will send us in circles
+                $isForeachCollection = false;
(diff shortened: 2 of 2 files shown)


felixfbecker:
There are merge conflicts, could you rebase?

felixfbecker:
As mentioned on Gitter, my thought on this is that if we try to resolve `$a` in `$a->getArray()`, the foreach statement node should not be traversed at all for finding the definition of `$a` (as any code in there will not have executed at the point where `$a->getArray()` is called). It should start at the previous sibling and then walk upwards the AST (parents and previous siblings), traversing into descendents as needed. Did you try that?

PR #568, opened 2017-12-28
fix(DefinitionResolver): find variables in sibling children

### LINKED ISSUE #566: PHP Intellisense, huge memory leak when building cache

felixfbecker:
Fixes #566

### DIFF
--- a/fixtures/completion/foreach.php
+++ b/fixtures/completion/foreach.php
@@ -38,3 +38,9 @@ foreach ($bar->test() as $value) {
 
 foreach ($unknownArray as $member->access => $unknown) {
     $unkno
+
+foreach ($loop as $loop) {
+}
+ 
+foreach ($loop->getArray() as $loop) {
+}
--- a/src/DefinitionResolver.php
+++ b/src/DefinitionResolver.php
@@ -545,6 +545,15 @@ class DefinitionResolver
         } else {
             throw new \InvalidArgumentException('$var must be Variable, Param or ClosureUse, not ' . get_class($var));
         }
+        if (empty($name)) {
+            return null;
+        }
+
+        $shouldDescend = function ($nodeToDescand) {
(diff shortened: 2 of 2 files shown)


phil-nelson:
Thanks, this is a better solution that mine, I was struggling to work out how to do something like this as your suggestion.

felixfbecker:
This also includes a change unrelated to the bug, in that before we only checked previous siblings and parents, but not the descendants of those. So this wouldn't register `$foo` before:
```php
if (true) {
  $foo = 123
}
$f| // no completion for $foo
```
now it should.

With that change I could make it so that the search for the foreach key/value starts at the parent/sibling too because it can then traverse into the ForeachStatement, but not if that was the starting point.

felixfbecker:
@jens1o @phil-nelson if anyone wants to test: [php-intellisense-1.5.4.vsix.zip](https://github.com/felixfbecker/php-language-server/files/1592295/php-intellisense-1.5.4.vsix.zip)

phil-nelson:
I've had a play with this branch, no problems so far :thumbsup:

jens1o:
Works liek a charm :)

Issue #372, opened 2017-05-02
IntelliSense always suggests "<?php" when in HTML


cstuder:
Every time I close an HTML tag in a PHP file, IntelliSense suggests `<?php` as next string. 

![image](https://cloud.githubusercontent.com/assets/288493/25624756/b4549210-2f5a-11e7-9585-f016aeb74076.png)

This is distracting when writing HTML: When I want to enter a linebreak, I first have to remove the suggestion before hitting `Enter`.

Is there a possibility to remove this particular suggestion?

(Using PHP IntelliSense V1.2.1 in Visual Studio Code 1.11.2 on Windows 10, `php.suggest.basic` is set to `false`)

felixfbecker:
This is because we need to set the `triggerCharacters` to `>` because a sequence like `->` is not supported. See https://github.com/Microsoft/language-server-protocol/issues/138

cstuder:
Just found the responsable code line: https://github.com/felixfbecker/php-language-server/blob/master/src/CompletionProvider.php#L281

I would prefer to have the first condition removed:

    } else if ($pos == new Position(0, 0)) {

My question is: Am I the only one who is annoyed by this? (Or am I the only one editing HTML in PHP files ?)

felixfbecker:
I don't understand how removing that condition would help

cstuder:
Sorry, I didn't make myself clear. Currently, the conditions are like this:

            } else if ($node instanceof Node\Stmt\InlineHTML || $pos == new Position(0, 0)) {

If I understand this correctly, the `<?php`-completion triggers on InlineHTML nodes and when the cursor is at the beginning of a document.

I propose to remove the first condition, so that line would only trigger on the second condition:

    } else if ($pos == new Position(0, 0)) {

Before opening a pull request I would like to understand why this line is how it is and whether my propsal is actually wanted.

felixfbecker:
The line is how it is to propose a new opening tag inside HTML. That is as designed. The issue is just that the autocomplete pop up all the time because we need to set `>` as a trigger character.

cstuder:
Yes, so before we go around in further circles: Do you think my request to remove this autocomplete is reasonable and that I should try my hands on a pull request to remove it?

felixfbecker:
I don't think so, that is as designed - Microsoft needs to fix the trigger characters

abeebola:
Hello.... I am having this issue as well and it is annoying to be honest. Is there a workaround or something?

[...]

PR #593, opened 2018-02-06
fix proposing php start tag when completion context is not given by the ls client

jens1o:
*cough* vscode *cough*

fixes https://github.com/felixfbecker/php-language-server/issues/372


jens1o:
Thanks!

PR #599, opened 2018-02-14
Autocomplete speedup vol. 2

Declspeck:
This PR does everything that the original Autocomplete Speedup (#451) does, and what your branch of it did:
- Refactors `Index` into a tree structure, rather than an array of Fqns to definitions.
- Use the new `Index` to speed up `::`, `->` completions.

In addition to that, this PR
- Also uses the new `Index` to speed up completion on names, e.g. `array_mer|` and adds benchmark for that.
- Rewrites name completion in terms of generators
- Makes the completion tests for name completion stricter (`TestCla|` should not complete `OtherClass`)
- Completes aliased namespaces, e.g.`use Microsof\PhpParser\Node; Node\Stat|` correctly, adds tests for this
- Only completes direct children, i.e. `TestNamesp|` no longer completes `TestNamespace\SubNamespace\Foo\Bar`. 

Misc. unrelated changes:
- Add documentation comments to `CompletionTest`
- Shorten lines in `CompletionProvider.php` to make it pass `composer lint`


felixfbecker:
It seems like there are a lot of changes in this PR that are unrelated to the rest, like adding more keywords. Could you cherry-pick the commits that are not strictly related to the autocomplete speedup out of this and submit them as individual PRs?

Declspeck:
Sure, maybe tomorrow. I'm not sure if the selectivity for `->` and `::` completions fixes a bug introduced by some of the earlier commits or if it fixes an issue that has always been there.

The relative specifier completion can of course be committed only on top of these changes, but I guess I can open a new PR after this has been merged.

Declspeck:
I just force-pushed (sorry about that - it seems like I cannot change the source branch of the PR) a new version with only the changes related to the speedup (i.e. rewrite of the global completion part.) The previous branch exists in [the branch autocomplet-speedup-copy](https://github.com/Declspeck/php-language-server/tree/autocomplet-speedup-copy) if anyone is interested. The following changes were saved for later:
- Autocomplete of relative names
- Autocomplete of `use`d functions and constants
- Added pseudo-keywords
- More selective completions at `Foo::something|` and `$foo->somethingEls|`

About the last part - for `Foo::something|`, `Foo::definitelyNotSomething` will also be completed. Is this intentional or a bug? It is even being tested for.

Declspeck:
I got another (unrelated) question about the completion logic - for roamed symbols, do we want to complete the name as fully qualified? It seems to make sense for `my_user_defined_fu|` -> `\my_user_defined_function`, but not for `array_mer|` -> `\array_merge`.

If so, it should probably be expressed as a `textEdit` rather than `insertText`, since at least `lsp-mode` in Emacs does not know what to do in the case where the insert text does not match the text that has been written.

Declspeck:
I added all the parts removed from this PR as separate PRs: #607 #606 #605 #604 #602

felixfbecker:
Currently when in a namespace we complete to `\array_merge` afaik. If we had settings there probably should be a setting for it.

felixfbecker:
Need some more time to review this.
Would you mind updating the description?

Declspeck:
Sure, I'll update it after work today (or at the very latest tomorrow) - I already forgot what I'm wanting to merge here, so I have to take another look myself. It looks like I also have a conflict to resolve.

Declspeck:
...And thanks for reviewing the other PRs!

Declspeck:
I updated the description and merged master into this.

src/CompletionProvider.php
```diff
@@ -280,114 +287,278 @@ public function provideCompletion(PhpDocument $doc, Position $pos, CompletionCon
...
+                $items = $this->getUnqualifiedCompletions($prefix, $currentNamespace, $importTables, $isCreation);
+            }
 
-            /** @var string[] $aliases A map from local alias to fully qualified name */
-            list($aliases,,) = $node->getImportTablesForCurrentScope();
+            $list->items = array_values(iterator_to_array($items));
+            foreach ($list->items as $item) {
+                // Remove ()
+                if (is_string($item->insertText) && substr($item->insertText, strlen($item->insertText) - 2) === '()') {
+                    $item->insertText = substr($item->insertText, 0, strlen($item->insertText) - 2);
```
jens1o: no need to use `strlen` here(this line and the line before).

Use `substr($item->insertText, -2);` instead.

src/Index/Index.php
```diff
@@ -216,10 +280,164 @@ public function unserialize($serialized)
     public function serialize()
     {
         return serialize([
-            'definitions' => $this->definitions,
+            'definitions' => iterator_to_array($this->getDefinitions(), true),
```
jens1o: the second parameter is already the default one, no need for the `true`.

Declspeck:
Thanks for the feedback @jens1o  - I finally got around to addressing it.

jens1o:
Thank you!

src/CompletionProvider.php
```diff
@@ -341,7 +341,7 @@ public function provideCompletion(
             foreach ($list->items as $item) {
                 // Remove ()
                 if (is_string($item->insertText) && substr($item->insertText, strlen($item->insertText) - 2) === '()') {
```
jens1o: this line can be simplified too

Tsine:
Hello
Do we have an estimation date on when this will be merged if it will be merged ?

felixfbecker:
Short answer, no, we don't have an estimate date.

felixfbecker:
Hey @Declspeck, sorry I had forgotten about this PR. Just took a look and it looks good to me. Do you think this is in a mergeable state / are you still interested in getting it across the finish line?

jens1o:
@felixfbecker I'd be interested in completing this PR. What do I need to do? Copy the clone?

felixfbecker:
Awesome. Here is roughly what you need to do:

Add the fork as a git remote:
```
git remote add Declspeck git@github.com:Declspeck/php-language-server.git
```
Fetch it:
```
git fetch Declspeck
```
Check it out:
```
git checkout autocomplet-speedup
```
Merge in master (make sure master is up to date):
```
git merge master
```
Fix merge conflicts and commit
Push to your own fork:
```
git push -u jens1o autocomplet-speedup
```

PR #601, opened 2018-02-15
Update travis php versions

vinkla:
Added PHP 7.1 and updated to stable version of 7.2.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,9 @@
-
 language: php
 
 php:
-  - '7.0'
-  - '7.2.0RC5'
+  - 7.0
+  - 7.1
+  - 7.2
 
 git:
   depth: 10
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,8 @@
 language: php
 
 php:
-  - 7.0
-  - 7.2
+  - '7.0'
+  - '7.2'
 
 git:
   depth: 10
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,6 @@ language: php
 
 php:
   - 7.0
-  - 7.1
   - 7.2
 
 git:


.travis.yml
```diff
@@ -1,9 +1,9 @@
-
 language: php
 
 php:
-  - '7.0'
-  - '7.2.0RC5'
+  - 7.0
+  - 7.1
```
felixfbecker: do we really need to test on 7.1 if we test the lowest version 7.0 and the newest version 7.2? Worried about build time for no gain

.travis.yml
```diff
@@ -1,9 +1,9 @@
-
 language: php
 
 php:
-  - '7.0'
-  - '7.2.0RC5'
+  - 7.0
```
felixfbecker: Please keep using strings. `7.0` == `7` == `7.x.x`, `'7.0'` == `7.0.x`

.travis.yml
vinkla: I've removed version 7.1 from the commit.

.travis.yml
vinkla: What? Didn't know about that. Do you know where I can find this in Travis documentation?

felixfbecker: I don't think that is a Travis thing. In YAML if you don't quote `7.0` it's a floating point number so it is the same as `7`. Now whatever version manager Travis uses to install runtimes will get `7`, not `7.0`, so specifying `7.0` is at least misleading. I know that `nvm install 7` will install the latest `7.x.x` version, while `nvm install 7.0` will install the latest `7.0.x` version. I don't know how it is for PHP but there must have been a reason for this commit https://github.com/felixfbecker/php-language-server/commit/41e84880b380368ae5862c01e9a00d55067c48ec and there is no disadvantage to using strings.

vinkla: Oh, I had no idea. Thanks for taking the time to explain.

PR #602, opened 2018-02-15
Completion for relative names

Declspeck:
This pull request is based on #599 so if it needs changes, I will have to port them to this PR too. I might just rebase and force-push this, so don't get too attached to the branch (e.g. don't branch from this).

This PR adds support for completion for relative names, i.e. completions at `namespace\|` and `namespace\Class\In\The\Current\Namesp|`


felixfbecker:
Is this still relevant?

PR #604, opened 2018-02-24
Add pseudo-keywords like int, bool, strict_types to completion

Declspeck:
This commit adds more pseudo-keywords to the completion.


Declspeck:
Updated the branch, so I guess this could be merged

PR #605, opened 2018-02-24
Complete class aliases whose targets do not exist

Declspeck:
This commit adds support and a test for the following case:

```php
use Some\Namespace\ClassWhichDoesNotExist;

ClassWhichDoesN| // Completes to ClassWhichDoesNotExist
```

This is primarily useful in cases where the target class is not indexed for some reason.

This PR is based on #599 so it cannot be merged before that and only the last commit f5b1256 contains new changes.


felixfbecker:
Is this still relevant?

PR #606, opened 2018-02-24
Autocomplete complete used functions and constants

Declspeck:
This add completion for 

```php
use function SomeNamespace\some_function;
use const SomeNamespace\SOME_CONST;

some_fun| // Completes to some_function
SOME_CONS| // Completes to SOME_CONST
```

This is based on #605 (Note - not the speedup PR, but another continuation of it), so this cannot be merged before #605 and #599 are merged.


felixfbecker:
Is this still relevant?

PR #607, opened 2018-02-24
Autocomplete member selectivity

Declspeck:
This removes completions for `TestClass::SOMETHING_ELSE` when completing for `TestClass::NOT_SOMETHING_ELS|`. Same for `->` completions.

This is based on #599 so this cannot be merged before that, and only the last commit contains any meaningful changes.


felixfbecker:
Is this still relevant?

PR #609, opened 2018-02-24
Refactor / performance: Track scope information while traversing

Declspeck:
This PR makes the resolution of variable types go forwards instead of backwards. It keeps an active scope. In my opinion, this makes the code more understandable (although the number of lines increased.) Code duplication is removed between completion for variables and resolving them in `DefinitionResolver`.

Additionally, the scope is used to cache results of `getResolvedName`, since it is quite slow, at least with the current Tolerant PHP Parser. This cache is invalidated when a namespace declaration is met.

There are no tests for `Scope` yet, in case there are severe issues with the general approach.

### DIFF
--- a/src/DefinitionResolver.php
+++ b/src/DefinitionResolver.php
@@ -277,10 +277,10 @@ class DefinitionResolver
             $name = $node->getName();
             // Resolve $this to the containing class definition.
             if ($name === 'this') {
-                if ($scope->currentClassLikeVariable === null) {
+                if ($scope->currentSelf === null) {
                     return null;
                 }
-                $fqn = substr((string)$scope->currentClassLikeVariable->type->getFqsen(), 1);
+                $fqn = substr((string)$scope->currentSelf->type->getFqsen(), 1);
                 return $this->index->getDefinition($fqn, false);
             }
 
@@ -297,16 +297,16 @@ class DefinitionResolver
         if ($fqn === 'self' || $fqn === 'static') {
             // Resolve self and static keywords to the containing class
             // (This is not 100% correct for static but better than nothing)
-            if ($scope->currentClassLikeVariable === null) {
+            if ($scope->currentSelf === null) {
                 return null;
             }
-            $fqn = substr((string)$scope->currentClassLikeVariable->type->getFqsen(), 1);
+            $fqn = substr((string)$scope->currentSelf->type->getFqsen(), 1);
         } else if ($fqn === 'parent') {
-            if ($scope->currentClassLikeVariable === null) {
+            if ($scope->currentSelf === null) {
                 return null;
             }
             // Resolve parent keyword to the base class FQN
-            $classNode = $scope->currentClassLikeVariable->definitionNode;
+            $classNode = $scope->currentSelf->definitionNode;
             if (!$classNode->classBaseClause || !$classNode->classBaseClause->baseClass) {
                 return null;
             }
@@ -439,10 +439,10 @@ class DefinitionResolver
             || $varType instanceof Types\Self_
         ) {
             // $this/static/self is resolved to the containing class
-            if ($scope->currentClassLikeVariable === null) {
+            if ($scope->currentSelf === null) {
                 return null;
             }
-            $classFqn = substr((string)$scope->currentClassLikeVariable->type->getFqsen(), 1);
+            $classFqn = substr((string)$scope->currentSelf->type->getFqsen(), 1);
         } else if (!($varType instanceof Types\Object_) || $varType->getFqsen() === null) {
             // Left-hand expression could not be resolved to a class
             return null;
@@ -498,7 +498,7 @@ class DefinitionResolver
 
         if ($className === 'self' || $className === 'static' || $className === 'parent') {
             // self and static are resolved to the containing class
-            $classNode = $scope->currentClassLikeVariable->definitionNode ?? null;
+            $classNode = $scope->currentSelf->definitionNode ?? null;
             if ($classNode === null) {
                 return null;
             }
@@ -509,7 +509,7 @@ class DefinitionResolver
                 }
                 $className = $scope->getResolvedName($classNode->extends);
             } else {
-                $className = substr((string)$scope->currentClassLikeVariable->type->getFqsen(), 1);
+                $className = substr((string)$scope->currentSelf->type->getFqsen(), 1);
             }
         } elseif ($scoped->scopeResolutionQualifier instanceof Node\QualifiedName) {
             $className = $scope->getResolvedName($scoped->scopeResolutionQualifier);
@@ -643,10 +643,10 @@ class DefinitionResolver
             }
             for ($i = 0; $t = $objType->get($i); $i++) {
                 if ($t instanceof Types\This) {
-                    if ($scope->currentClassLikeVariable === null) {
+                    if ($scope->currentSelf === null) {
                         return new Types\Mixed_;
                     }
-                    $classFqn = substr((string)$scope->currentClassLikeVariable->type->getFqsen(), 1);
+                    $classFqn = substr((string)$scope->currentSelf->type->getFqsen(), 1);
                 } else if (!($t instanceof Types\Object_) || $t->getFqsen() === null) {
                     return new Types\Mixed_;
                 } else {
@@ -920,15 +920,15 @@ class DefinitionResolver
         $className = $scope->getResolvedName($class);
 
         if ($className === 'self') {
-            if ($scope->currentClassLikeVariable === null) {
+            if ($scope->currentSelf === null) {
                 return new Types\Self_;
             }
-            return $scope->currentClassLikeVariable->type;
+            return $scope->currentSelf->type;
         } else if ($className === 'parent') {
-            if ($scope->currentClassLikeVariable === null) {
+            if ($scope->currentSelf === null) {
                 return new Types\Object_;
             }
-            $classNode = $scope->currentClassLikeVariable->definitionNode;
+            $classNode = $scope->currentSelf->definitionNode;
             if (empty($classNode->classBaseClause)
                 || !$classNode->classBaseClause->baseClass instanceof Node\QualifiedName
             ) {
@@ -1027,8 +1027,8 @@ class DefinitionResolver
                 && ($returnType = $returnTags[0]->getType()) !== null
             ) {
                 // Use @return tag
-                if ($returnType instanceof Types\Self_ && null !== $scope->currentClassLikeVariable) {
-                    return $scope->currentClassLikeVariable->type;
+                if ($returnType instanceof Types\Self_ && null !== $scope->currentSelf) {
+                    return $scope->currentSelf->type;
                 }
                 return $returnType;
             }
@@ -1037,8 +1037,8 @@ class DefinitionResolver
                 if ($node->returnType instanceof PhpParser\Token) {
                     // Resolve a string like "bool" to a type object
                     return $this->typeResolver->resolve($node->returnType->getText($node->getFileContents()));
-                } else if ($scope->currentClassLikeVariable !== null && $scope->getResolvedName($node->returnType) === 'self') {
-                    return $scope->currentClassLikeVariable->type;
+                } else if ($scope->currentSelf !== null && $scope->getResolvedName($node->returnType) === 'self') {
+                    return $scope->currentSelf->type;
                 }
                 return new Types\Object_(new Fqsen('\\' . $scope->getResolvedName($node->returnType)));
             }
@@ -1160,10 +1160,10 @@ class DefinitionResolver
         // }
         if ($node instanceof Node\MethodDeclaration) {
             // Class method: use ClassName->methodName() as name
-            if ($scope->currentClassLikeVariable === null) {
+            if ($scope->currentSelf === null) {
                 return;
             }
-            $className = substr((string)$scope->currentClassLikeVariable->type->getFqsen(), 1);
+            $className = substr((string)$scope->currentSelf->type->getFqsen(), 1);
             if (!$className) {
                 // Ignore anonymous classes
                 return null;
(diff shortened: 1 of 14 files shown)


src/Scope/TreeTraverser.php
```diff
@@ -0,0 +1,193 @@
...
+                        $use
+                    );
+                }
+            }
+
+            return $childScope;
+        }
+
+        if ($node instanceof ClassLike
+            && (in_array($childName, ['classMembers', 'interfaceMembers','traitMembers'], true))
```
jens1o: Do we need to `strictly` compare here?

src/Scope/TreeTraverser.php
```diff
@@ -0,0 +1,193 @@
...
+                $node
+            );
+        } else if ($node instanceof Node\Statement\NamespaceDefinition) {
+            // After a new namespace A\B;, the current alias table is flushed.
+            $scope->clearResolvedNameCache();
+        }
+
+
+        // TODO: Handle use (&$x) when $x is not defined in scope.
+        // TODO: Handle list(...) = $a;
```
jens1o: and PHP7+ `[...] = $a;`

src/Scope/TreeTraverser.php
Declspeck: Not really - I thought it would be good form, just like using `===` instead of `==` - do you think there's a reason to change it?

src/Scope/TreeTraverser.php
Declspeck: :+1: I didn't even know about that construct

Declspeck: Ok, I added a TODO comment for that and `foreach ($a as [...])`

src/Scope/TreeTraverser.php
jens1o: Somehow my benchmarks are showing different results:
https://gist.github.com/jens1o/621c1307ee0b9d618839211688d46dba

```php
PS C:\nginx\html\projects\php7.2-playground> php .\in_array_strict_benchmark.php
Took 0.42s with strict compare
Took 0.35s without strict compare
Took 5.47s with strict compare
Took 0.43s without strict compare
Took 0.54s with strict compare
Took 0.47s without strict compare
```

staabm: Your benchmark shows searching for an empty string which does not is the case here?

staabm: @nikic is it expexcted that strictly searching for any empty string via in_array  is slower then when doing it non-strict?

jens1o: > Your benchmark shows searching for an empty string which does not is the case here?

I didn't knew what I could use there.

nikic: @staabm Well, that *specific* benchmark is rather useless, as the second array contains 0 as first element, which is (non-strict) equal to the empty string, so what you see there is the difference between matching on the first element in the array vs scanning through all 10000 elements and not finding anything.

nikic: That said, the in_array strict path is slightly less optimized than the non-strict one. Though if you want to micro-optimize this, the way to do it is to use `\in_array` (with strict), which will transform to an HT lookup. Or even better, just directly implement it as an HT lookup so you're not dependent on optimization behavior.

Declspeck: Ah, I didn't actually mean performance-wise but style-wise.

I think it's better because the code does not perform convoluted type casts under the hood. It's not that the code would be more incorrect that way, just that it *feels* that way - when I'm searching for a bug, I don't have to wonder if this comparison here does something funny.

Declspeck: ... although I would have expected the strict version also to be faster.

jens1o: > Well, that specific benchmark is rather useless, as the second array contains 0 as first element, which is (non-strict) equal to the empty string, so what you see there is the difference between matching on the first element in the array vs scanning through all 10000 elements and not finding anything.

Good point!

src/DefinitionResolver.php
```diff
@@ -716,6 +603,7 @@ public function resolveExpressionNodeToType($expr)
             if ($token === PhpParser\TokenKind::NullReservedWord) {
                 return new Types\Null_;
             }
+            return new Types\Mixed_;
```
felixfbecker: Do you expect this case to ever get hit? Or is it for future-proofing?

src/Scope/GetScopeAtNode.php
```diff
@@ -0,0 +1,57 @@
...
+    $sourceFile = $targetNode->getRoot();
+    /** @var FunctionDeclaration|null The first function declaration met, excluding anonymous functions. */
+    $nearestFunctionDeclarationParent = $targetNode->getFirstAncestor(FunctionDeclaration::class);
+    /** @var ClassLike|null The first class met. */
+    $nearestClassLike = $targetNode->getFirstAncestor(ClassLike::class);
+
+    $traverser = new TreeTraverser($definitionResolver);
+    $resultScope = null;
+    $traverser->traverse(
+        $sourceFile,
```
felixfbecker: I don't really understand why you would want to traverse the whole source file from the root if you already have the node and every node is linked to its parent. This seems very inefficient, as you will visit a lot of nodes that you don't care about. A scope is intuitively definded as the the "area

src/Scope/Scope.php
```diff
@@ -0,0 +1,45 @@
...
+
+use Microsoft\PhpParser\Node\QualifiedName;
+
+/**
+ * Contains information about variables at a point.
+ */
+class Scope
+{
+    /**
+     * @var Variable|null "Variable" representing this/self
```
felixfbecker: ...or `null` if the scope is not inside a class

src/Scope/Scope.php
```diff
@@ -0,0 +1,45 @@
+<?php
+declare(strict_types=1);
+
+namespace LanguageServer\Scope;
+
+use Microsoft\PhpParser\Node\QualifiedName;
+
+/**
+ * Contains information about variables at a point.
```
felixfbecker: Could you elaborate what "point" means here? How is the scope boundary defined?

src/Scope/Scope.php
```diff
@@ -0,0 +1,45 @@
...
+        $this->resolvedNameCache = [];
+    }
+
+    /**
+     * @return string|null
+     */
+    public function getResolvedName(QualifiedName $name)
+    {
+        $nameStr = (string)$name;
+        if (array_key_exists($nameStr, $this->resolvedNameCache)) {
```
felixfbecker: I am assuming there are no `null` values in the array so `isset` would be better

src/Scope/TreeTraverser.php
```diff
@@ -0,0 +1,193 @@
...
+use Microsoft\PhpParser\Node\Expression;
+use Microsoft\PhpParser\Token;
+use Microsoft\PhpParser\TokenKind;
+use phpDocumentor\Reflection\Fqsen;
+use phpDocumentor\Reflection\Types;
+
+/**
+ * Traversers AST with Scope information.
+ */
+class TreeTraverser
```
felixfbecker: The parser already exposes utilities to traverse the tree (and uses the iterator pattern instead of the visitor pattern, which is superior imo). Given that this is a fair amount of code, could you explain why this is needed? What is the difference between this scope-tree traverser and a general traverser?

src/Scope/TreeTraverser.php
felixfbecker: I agree that strict comparison is always easier to reason about.

src/DefinitionResolver.php
```diff
@@ -175,10 +177,15 @@ private function getDocBlock(Node $node)
      *
      * @param Node $node
      * @param string $fqn
+     * @param Scope|null $scope Scope at the point of Node. If not provided, will be computed from $node.
```
felixfbecker: How do you ensure that the scope passed in is the scope of the node?

src/DefinitionResolver.php
Declspeck: These were added accidentally (same goes the `Types\Mixed_` below) - I tried various things in the same branch if they improved performance, including bailing early here. It did not really affect performance, but I apparently forgot to revert some changes. I can revert this.

src/Scope/Scope.php
Declspeck: Scope when evaluating the expression at the node being traversed, if it were an expression.

... perhaps? I'm not sure how to express it clearly.

src/Scope/TreeTraverser.php
Declspeck: I need to look into the parser code and see if we could use it - my initial guess is that we can't though, since the traversing needs to happen in the following fashion:

```
Scope = []                      //    Scope 1 - Root
    Scope = [$this]             //    Scope 2 - Class entered 
        Scope = [$this, $a]     //    Scope 3 - Function entered
    Scope = [$this]             // !! Scope 2 - Same scope again 
Scope = []                      // !! Scope 1 - Same scope again
```

The lines marked with `!!` - if we only get a callback per `Node` or `Token`, we cannot know when a scope is exited. Additionally, we'd need to manage a stack manually instead of relying on the call stack.

Further, we need information on the context of the element - if we have a `BracedStatementList` (or whatever it's called), we need to know if it is a `functionBodyOrSemicolon` to start a new `Scope`, so the `Node` alone is not sufficient.

If the parser does not provide a traverser that fits those two needs, I think that working around them might make the code more confusing.

src/DefinitionResolver.php
Declspeck: I have to think about this a bit more.

One thing that comes to mind is that the `Scope` could contain information about the `Node` or `FunctionLike` it is defined at? The added check would have a performance penalty (which I guess might be significant), since you'd need to find the closest `FunctionLike`, `ClassLike`, or `SourceFileNode` ancestor every time you wanted to check if the scope belongs to a node.

src/Scope/Scope.php
Declspeck: True, I'll fix it. I just copy-pasted code from the DocBlock thingie.

src/DefinitionResolver.php
```diff
@@ -726,6 +614,7 @@ public function resolveExpressionNodeToType($expr)
             if ($def !== null) {
                 return $def->type;
             }
+            return new Types\Mixed_;
```
Declspeck: @felixfbecker This is not related to this PR, I left it here by accident. Should I revert it or let it stay?

src/Scope/TreeTraverser.php
Jasu: I looked into this, and  `getDescendantNodesAndTokens` or `getDescendantNodes` cannot really be used, since we won't know when a node is exited and the scope should be restored.

I think that `TreeTraverser` could be converted into a generator yielding something - e.g. rename `Scope` to `TraversingContext` which contains both `$node` and `$variables` etc. Do you think that's worth pursuing?

Declspeck: I looked into this, and `getDescendantNodesAndTokens` or `getDescendantNodes` cannot really be used, since we won't know when a node is exited and the scope should be restored.

I think that `TreeTraverser` could be converted into a generator yielding something - e.g. rename `Scope` to `TraversingContext` which contains both `$node` and `$variables` etc. Do you think that's worth pursuing?

(I accidentally posted this with a wrong GitHub account first - sorry about the noise)

felixfbecker: So first of all I still think parent traversal is the better way to go.
And then `getDescendentNodes` afaik takes a callback to decide whether a node should be entered or not.
For traversal that needs full control over the recursive aspect I did a PR a while ago to tolerant-php-parser with RecursiveIterator support: https://github.com/Microsoft/tolerant-php-parser/pull/139
I didn't need it at the time though and it turned out it was a tad slower than the generators. It would be interesting to see how it compares to this TreeTraverser and the previous implementation, especially on latest PHP 7.2.

Declspeck: The `RecursiveIterator` thing certainly looks interesting. It would require manual handling of a stack though, if I understand it correctly.

I think that traversing the code backwards to find types has a few problems, which I tried to solve with this:

#### 1. Duplicated work:

Find references on line 2:
```php
1 $a = new A;     // 2.                    Ok, `$a is A`
2 $b = $a->foo(); // 1. Start traversing - need to find what $a is
3 $b->bar();
```
After that, find references on line 3:
```php
1 $a = new A;     // 3. Duplicated work  - Ok, `$a is A`
2 $b = $a->foo(); // 2. Duplicated work  - Need to find out what $a is
3 $b->bar();      // 1. Start Traversing - need to find what $b is to get reference to bar
```
This could of course be solved with some sort of memoization. It would probably end up building a scope, but in reverse order.

#### 2. Slowness of parsing backwards in Tolerant PHP Parser
Getting a previous sibling gets its parent, enumerates its children, and when it finds the original node, it returns the previously met one.

#### (3. Harder to understand - this is mostly a matter of preference maybe)
Since code is evaluated from top to bottom, an approach evaluating code from top to bottom is easier to understand, at least for me.

*I'm going to sleep now, so I won't reply for a while. I might take a look at the other PRs and comments tomorrow.*

felixfbecker: > 1. Duplicated work:

I don't think this problem has anything to do with how to traverse. The difference is that this PR introduces a Scope object, where before this was always computed JIT. I never had any performance problems with find-refs on variables though because their scope is naturally small in PHP, so it's a CPU vs memory usage trade off.

> 2. Slowness of parsing backwards in Tolerant PHP Parser
> Getting a previous sibling gets its parent, enumerates its children, and when it finds the original node, it returns the previously met one.

This is a good point. With PHPParser I added the siblings as properties so that was fast. We could do the same for tolerant-parser.

----------

Are you maintaining the same behaviour as before with the new traverser?

Declspeck: >  I never had any performance problems with find-refs on variables though because their scope is naturally small in PHP, so it's a CPU vs memory usage trade off.

It might be that the performance benefits here come from caching `getResolvedName` results in the `Scope` object. I started working on this because XDebug profiling showed finding variable types as a significant cost. XDebug has very high overhead though, so it might be completely inaccurate. It might be worth trying to cache the `getResolvedName` results alone, especially if you don't like the `Scope` stuff.

> Are you maintaining the same behaviour as before with the new traverser?

I believe so - the variables in `Scope` are overwritten when a new assignment is met, so on line 3 `$a` would be a string.

Issue #280, opened 2017-02-04
array with strings gets indexed as int[]


jens1o:
![image](https://cloud.githubusercontent.com/assets/11234139/22621356/1f380fbc-eb22-11e6-97b0-f1ccbdd46473.png)

Snippet: 
```php
<?php

$someArray = ['Hello World', 'Bye'];

$some|
```

jens1o:
After doing some research, I've lifted up here https://github.com/phpDocumentor/TypeResolver/blob/master/src/Types/Array_.php#L49 (but it's final)... We need to check every single one to be sure it's right...

I'm proposing adding the label "uplift"

felixfbecker:
Sorry, I'm not following. What are you proposing?
The bug must be in `DefinitionResolver->resolveExpressionNodeToType()`

PR #610, opened 2018-02-25
Fix: Allow integer FQNs in references in PhpDocument

Declspeck:
Sometimes with incorrect code, a number is parsed as an FQN. When it is added to an array as its key, it will be converted to an integer. This causes a type error when it is added to references.

This fixes an exception that gets thrown when running Performance.php. I also removed the `try { ... } catch` block from Performance.php, which would make testing of performance changes easier as increased exception rate would not be interpreted as increased performance.

### DIFF
--- a/Performance.php
+++ b/Performance.php
@@ -51,11 +51,7 @@ foreach($frameworks as $framework) {
         $definitionResolver = new DefinitionResolver($index);
         $parser = new PhpParser\Parser();
 
-        try {
-            $document = new PhpDocument($testCaseFile, $fileContents, $index, $parser, $docBlockFactory, $definitionResolver);
-        } catch (\Throwable $e) {
-            continue;
-        }
+        $document = new PhpDocument($testCaseFile, $fileContents, $index, $parser, $docBlockFactory, $definitionResolver);
     }
 
     echo "------------------------------\n";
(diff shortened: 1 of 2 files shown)


jens1o:
But this does not fix https://github.com/felixfbecker/php-language-server/issues/280, does it?

Declspeck:
@jens1o No, this is a separate issue. I'm not exactly sure what triggers this, but I imagine that if php-language-server sees something like `new 2;`, it will throw an exception. I don't know what the exact incorrect code being parsed needs to be to make PhpDocument throw.

jens1o:
Ah, now I get it. Thank you for elaborating on this and the fix :)

Performance.php
```diff
@@ -51,11 +51,7 @@
         $definitionResolver = new DefinitionResolver($index);
         $parser = new PhpParser\Parser();
 
-        try {
-            $document = new PhpDocument($testCaseFile, $fileContents, $index, $parser, $docBlockFactory, $definitionResolver);
-        } catch (\Throwable $e) {
-            continue;
-        }
+        $document = new PhpDocument($testCaseFile, $fileContents, $index, $parser, $docBlockFactory, $definitionResolver);
```
felixfbecker: Does the benchmark still run without this on all frameworks? I think I added this cause some files are invalid (on purpose, e.g. test fixtures) and this would break the whole loop

felixfbecker:
Fix LGTM

Performance.php
Declspeck: It does indeed. The `(string)` thing fixes the test fixture thing in `math-php`. I also have (somehow by accident) downloaded Tolerant PHP Parser's validation frameworks, and it doesn't choke on them either.

PR #612, opened 2018-02-28
Remove unused 'use' statements

TysonAndre:
detected via static analysis and manually checked.

Unit tests continue to pass.

### DIFF
--- a/src/Cache/FileSystemCache.php
+++ b/src/Cache/FileSystemCache.php
@@ -3,7 +3,6 @@ declare(strict_types = 1);
 
 namespace LanguageServer\Cache;
 
-use LanguageServer\LanguageClient;
 use Sabre\Event\Promise;
 
 /**
--- a/src/Client/TextDocument.php
+++ b/src/Client/TextDocument.php
@@ -4,7 +4,7 @@ declare(strict_types = 1);
 namespace LanguageServer\Client;
 
 use LanguageServer\ClientHandler;
-use LanguageServer\Protocol\{Message, TextDocumentItem, TextDocumentIdentifier};
+use LanguageServer\Protocol\{TextDocumentItem, TextDocumentIdentifier};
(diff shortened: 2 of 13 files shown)


felixfbecker:
Is there a way to check this in CI? These changes will otherwise erode quickly

TysonAndre:
When you asked initially, there wasn't a way to check this in CI - There were about 10 false positives due to inline `@var` annotations, which didn't parse inline var annotations.

It's definitely possible now thanks to a third party plugin that parses inline var annotations: See https://github.com/felixfbecker/php-language-server/compare/master...TysonAndre:add-phan-config?expand=1 for a Phan config for that . I'd recommend using phan.phar from the releases in travis (easiest) or cloning and installing phan in a separate directory - Phan also depends on tolerant-php-parser, but would hold it back to an older version.

- See https://github.com/phan/phan/wiki/Getting-Started#running-phan-in-continuous-integration (Instructions would be similar, but I recommend `path/to/phan.phar` instead of `vendor/bin/phan` here)

The errors it would emit are https://gist.github.com/TysonAndre/516356589be4230694952088b84eb9b6 . See https://github.com/phan/phan/wiki/Incrementally-Strengthening-Analysis#suppressing-issues (Adding pre-existing issues to suppress_issue_list is one way to get started)

felixfbecker:
@TysonAndre would you do a PR to add this to CI? 🙌

TysonAndre:
How do you want that check set up? A whitelist just for unused use statements? A blacklist for issues with high false positives?

felixfbecker:
I like linting and typechecking to be as strict as possible. I'd enable everything that are obvious mistakes (or can be easily fixed by rewriting code) and everything that is a mistake in almost all cases (and then ignore errors case-by-case with an ignore comment if needed and if Phan supports that). But really any checks are better than no checks.

PR #618, opened 2018-03-12
added a TOC to the README

staabm:
as the README already is super long this would bring some help for first time users.

generated with http://tableofcontents.herokuapp.com/


felixfbecker:
The ToC looks broken for me:

image

But besides that, frankly I think listing all the features in the ToC just to link to the screenshots bloats the README more than it simplifies it.

staabm:
I simplified the TOC so the main points are mentioned.
do you like it like that?

PR #625, opened 2018-03-23
Remove unused use statements, nit on phpdoc

TysonAndre:
Add a note on something that looks like an invalid array index

Also, see the note on https://github.com/felixfbecker/php-language-server/pull/612#issuecomment-375547570 (This issue (and other issues) are possible to detect in CI, even for uses in inline doc comments such as `/** @var T */ $x = some_expr();`)


src/Cache/ClientCache.php
```diff
@@ -11,6 +11,8 @@
  */
 class ClientCache implements Cache
 {
+    public $client;
```
felixfbecker: Could you add a docblock with an `@var` type annotation?

Issue #105, opened 2016-10-21
Avoid prepending <?php in hover messages


kaloyan-raev:
At the moment the first `MarkedString` in hover messages is of language "php". According to LSP specification the client should render this as if it was a markdown string like:
<pre>```php
php code block
```</pre>
The language server prepends `<?php` to ensure that the VSCode colorize the PHP code. 

I haven't seen any markdown editor to require wrapping PHP code in PHP tags to colorize it. It seems to be a specific implementation detail in VSCode. For all other clients this `<?php` tag does not bring any information to users, it just stands there like a buggy code hickup.

I propose one of the following solution:
- Don't prepend `<?php` at all and report the issue to VSCode.
- Prepend `<?php` only if the client is VSCode.

felixfbecker:
If you think about it, VS Code is correct here. Any code outside of `<?php` tags in PHP is HTML, and should be colorized like HTML. I met this issue before in my debug adapter. The only semantically-correct solution would be to do have a separate language ID for the PHP _code_, like `phpcode`.

kaloyan-raev:
> Any code outside of <?php tags in PHP is HTML, and should be colorized like HTML.

Exactly! (Although the type of the text outside the PHP tags depends on the HTTP response content type). So only the code _inside_ the PHP tags is PHP code, not the PHP tags themselves. PHP tags are necessary for the PHP interpreter to know where the PHP code starts and finishes. However, the Markdown viewer/editor does not need them, because it has \`\`\`php and \`\`\` for this purpose. The text between \`\`\`php and \`\`\` is only PHP code and should be treated like this.

If you really believe that VSCode is the ultimate source of truth for Markdown then I suggest to go for the second proposed solution:
- Prepend `<?php` only if the client is VSCode.

felixfbecker:
Hm, I see your point. In a CLI, anything outside `<?php` will actually just get printed to STDOUT and doesn't have to be HTML.

I don't want to implement any client-dependent behaviour. The LS must function universal for every client. Besides, I don't even know how to find out if it's VS Code who launched us.

kaloyan-raev:
Eclipse Che adds `"clientName": "EclipseChe"` to `InitializeParams` when sending the Initialize Request. This additional param is actually provided by the Typefox Java binding and is a custom extension of LSP. We can propose it becomes part of the protocol.

felixfbecker:
[...]

PR #631, opened 2018-04-06
Don t prepend returned hover string with `<?php`

WouterSioen:
Fixes https://github.com/felixfbecker/php-language-server/issues/105

This can be removed since VS Code changed their implementation according to https://github.com/Microsoft/vscode/issues/14166#issuecomment-275600794

### DIFF
--- a/src/Server/TextDocument.php
+++ b/src/Server/TextDocument.php
@@ -338,7 +338,7 @@ class TextDocument
             }
             $contents = [];
             if ($def->declarationLine) {
-                $contents[] = new MarkedString('php', "<?php\n" . $def->declarationLine);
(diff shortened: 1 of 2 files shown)


felixfbecker:
I don't see any recent activity on the issue that you linked and it is still open (note that I filed it). Did you test this? If you posted a screenshot proofing that this works I could merge this

WouterSioen:
The comment looks like it’s fixed since January, but I don’t use VS Code so I’m not sure at the moment. I’ll test after the weekend.

felixfbecker:
The comment you linked just says they are removing it from the iteration plan for January (2017) and moving it back to Backlog

PR #636, opened 2018-04-19
Update .travis.yml

andreybolonin:


felixfbecker:
What is the reason for this change? The amount of versions to test is a trade off between build time and safety. What could break in 7.1 that would not break in 7.0?

PR #639, opened 2018-04-26
Include local and global installation instructions in README.md.

LeviHarman:
This pull request alters the installation section of the readme to include the installation instructions found at [emacs-lsp/lsp-php](https://github.com/emacs-lsp/lsp-php )

I found that guide helpful. Hopefully it helps others.

### DIFF
--- a/README.md
+++ b/README.md
@@ -133,18 +133,9 @@ All classes are considered internal and are not subject to semver.
 ## Installation
 
 The recommended installation method is through [Composer](https://getcomposer.org/).
-Simply run
 
-    composer require felixfbecker/language-server
 
-and you will get the latest stable release and all dependencies.  
-Running `composer update` will update the server to the latest non-breaking version.
-
-After installing the language server and its dependencies,
-you must parse the stubs for standard PHP symbols and save the index for fast initialization.
-
-     composer run-script --working-dir=vendor/felixfbecker/language-server parse-stubs
-# Local installation
+### Local Installation
 Create a directory for php-language-server. Create a composer.json file in it, with the following contents:
 
 ```
@@ -176,7 +167,15 @@ composer global require felixfbecker/language-server
 composer global run-script --working-dir=vendor/felixfbecker/language-server parse-stubs
 ```
 
-Slightly altered from a guide found [here ](https://github.com/emacs-lsp/lsp-php). 
+
+and you will get the latest stable release and all dependencies.  
+Running `composer update` will update the server to the latest non-breaking version.
+
+After installing the language server and its dependencies,
+you must parse the stubs for standard PHP symbols and save the index for fast initialization.
+
+     composer run-script --working-dir=vendor/felixfbecker/language-server parse-stubs
+
 ## Running
 
 Start the language server with
--- a/README.md
+++ b/README.md
@@ -134,6 +134,7 @@ All classes are considered internal and are not subject to semver.
 
 The recommended installation method is through [Composer](https://getcomposer.org/).
 
+```composer require felixfbecker/language-server```
 
 ### Local Installation
 Create a directory for php-language-server. Create a composer.json file in it, with the following contents:
--- a/README.md
+++ b/README.md
@@ -140,14 +140,21 @@ Create a directory for php-language-server. Create a composer.json file in it, w
 ```
 {
     "minimum-stability": "dev",
-    "prefer-stable": true
+    "prefer-stable": true,
+    "require": {
+        "felixfbecker/language-server": "v5.4.0"
+    },
+    "scripts": {
+        "parse-stubs": "LanguageServer\\ComposerScripts::parseStubs",
+        "post-update-cmd": "@parse-stubs",
+        "post-install-cmd": "@parse-stubs"
+    }
 }
 ```
 Then, in the directory, run the following commands:
 
 ```
 composer require felixfbecker/language-server
-composer run-script --working-dir=vendor/felixfbecker/language-server parse-stubs
 ```
 
 ### Global installation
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -132,18 +132,39 @@ All classes are considered internal and are not subject to semver.
...
-     composer run-script --working-dir=vendor/felixfbecker/language-server parse-stubs
+The recommended installation method is through [Composer](https://getcomposer.org/).  Follow the following installation instructions for local or global installation after installing composer on your system.
+
+### Local Installation
+Create a directory for php-language-server. Create a composer.json file in it, with the following contents:
+
+```
+{
+    "minimum-stability": "dev",
+    "prefer-stable": true
```
felixfbecker: I would recommend adding these lines:

https://github.com/felixfbecker/vscode-php-intellisense/blob/21d79b5fd64fb7dcc09d97febf48ade309256e65/composer.json#L7-L11

so then it's not needed to run manually.

LeviHarman: Okay I changed it. I am unsure at the moment if I need the following line.
```
After editing your composer.json, you can install felixfbecker/php-language-server.
composer global require felixfbecker/language-server
````

I can check if the line above is needed but it will have to be tomorrow. I'm about to turn off the computer for the day.

Comment (LeviHarman) [reply README.md]: I checked the lines in question above they are needed.

**Comment by felixfbecker** [comment]
Side question, what is the use case for installing globally?

**Comment by LeviHarman** [comment]
I haven't done it myself. I included it because it was in emacs-lsp/lsp-php readme. I'll ask the creator of the README and see why he included it.

**Comment by cixtor** [comment]
Hello @LeviHarman did you hear back from emacs-lsp?

I wonder if you can just remove the global installation instructions from this pull-request so it can be merged and create a new pull-request later when the maintainers of the other LSP project answer your question about the usability of having this project globally accessible. I recently got bitten by issue #611 after attempting to have SublimeText + Dockerized (PHP-LSP) and your explanation [here](https://github.com/felixfbecker/php-language-server/issues/611#issuecomment-384709284) was the only thing that got me back on track.

It would be good to have the local installation instructions merged so others can benefit from it.

**Comment by rchl** [comment]
Would appreciate this getting merged too as current instructions in the readme did not work for me and I wasn't sure why as I don't know composer.

I got this error with default instructions:
`% composer require felixfbecker/language-server
Using version ^5.4 for felixfbecker/language-server
./composer.json has been created
Loading composer repositories with package information
Updating dependencies (including require-dev)
Your requirements could not be resolved to an installable set of packages.

  Problem 1
    - felixfbecker/language-server v5.4.2 requires jetbrains/phpstorm-stubs dev-master -> satisfiable by jetbrains/phpstorm-stubs[dev-master] but these conflict with your requirements or minimum-stability.
    - felixfbecker/language-server v5.4.1 requires jetbrains/phpstorm-stubs dev-master -> satisfiable by jetbrains/phpstorm-stubs[dev-master] but these conflict with your requirements or minimum-stability.
    - felixfbecker/language-server v5.4.0 requires jetbrains/phpstorm-stubs dev-master -> satisfiable by jetbrains/phpstorm-stubs[dev-master] but these conflict with your requirements or minimum-stability.
    - Installation request for felixfbecker/language-server ^5.4 -> satisfiable by felixfbecker/language-server[v5.4.0, v5.4.1, v5.4.2].


Installation failed, deleting ./composer.json.
`

**Comment by rchl** [comment]
Well, to be fair these instructions didn't help either. Still getting same error.

But personally I think installing globally makes sense as this is the kind of package that IMO should be installed globally in user's directory rather than in specific project as it is generally consumed by text editors.

**Comment by rchl** [comment]
I have found #611 and managed to fix my issue.
Problem for me was that I already had `~/.composer` directory so running `composer global require felixfbecker/language-server` used composer.json from that directory instead of global one from `~/.config/composer/composer.json`.

File: README.md [hunk 2]
```diff
@@ -132,18 +132,54 @@ All classes are considered internal and are not subject to semver.
...
+}
+```
+Then, in the directory, run the following commands:
+
+```
+composer require felixfbecker/language-server
+```
+
+### Global installation
+Before installing php-language-server, make sure your ~/.config/composer/composer.json includes the lines below. The settings apply to all globally installed Composer packages, so proceed with caution. If you do not want to edit your global Composer configuration, see the section for local installation above.
```
jason0x43, line 161: At least on my system (Mac with homebrew installed php and composer), the global config is in `~/.composer/composer.json` rather than `~/.config/composer/composer.json`.

phil-s: More generally, use `$COMPOSER_HOME` if it's set.  Refer to https://getcomposer.org/doc/03-cli.md#composer-home

If it's not set, note that a hard-coded fallback could be error-prone, as the default value is quite variable depending on the OS (and version thereof) you're running.

README.md
```diff
@@ -132,18 +132,54 @@ All classes are considered internal and are not subject to semver.
...
-Running `composer update` will update the server to the latest non-breaking version.
-
-After installing the language server and its dependencies,
-you must parse the stubs for standard PHP symbols and save the index for fast initialization.
-
-     composer run-script --working-dir=vendor/felixfbecker/language-server parse-stubs
+The recommended installation method is through [Composer](https://getcomposer.org/).  Follow the following installation instructions for local or global installation after installing composer on your system.
+
+### Local Installation
+Create a directory for php-language-server. Create a composer.json file in it, with the following contents:
```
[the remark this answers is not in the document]
phil-s: I believe the idea here is that some users will want to try this out without making any changes to their global composer config, and these instructions provide a way of creating a local (perhaps temporary) instance of the language server without touching anything else.

The text editor can then be pointed to the local installation for testing.  If it all pans out, the user might move to a global install (or not, as the case may be).

So "local" installation here can be thought of as a self-contained installation, rather than integrating it into another project like vscode-php-intellisense.  Perhaps that should be covered as a third case, though?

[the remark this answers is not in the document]
phil-s: An example of why users might want a separate/self-contained installation is so that they can patch the language server for specific use-cases without affecting their global installation.  For instance, modifying the hard-coded set of filename extensions which should be treated as PHP files in a particular project type -- a need that I personally ran into yesterday.

PR #509, opened 2017-10-30
WIP: multi-root support

felixfbecker:
closes #513

Issue #465, opened 2017-08-22
Cpu load, memory leak


pakulin:
mac sierra
Code 1.15.1
Php instellisense 1.15.0
php 7.1.6_18

Cpu load 95% with no tabs opened, memory usage gradually increased by 1-2 mb in 10 secs
Maybe this will help: 
[
[php_intellisense_1.5.0.txt](https://github.com/felixfbecker/php-language-server/files/1240892/php_intellisense_1.5.0.txt)
](url)

venatiodecorus:
I also see a PHP process start and utilize a full core of my CPU as soon as I open VSCode.

OSX 10.12.5
VSCode 1.15.1
PHP Intellisense 1.5.1

alancd:
"All 8812 PHP files parsed in 492 seconds. 160 MiB allocated"

felixfbecker:
What is the bug here? The language server needs to build it's index the first time when it's not cached, it's expected that this takes some CPU and RAM

alancd:
It looks like there's no caching, and every time you open a project with VS CODE, a parsing process is opened.

felixfbecker:
dependencies are cached

coderNeos:
+1 i have the same problem. Everytime i relaunch Vscode, parsing starts from begining and dependencies are not cached - it may be becayse i have 2 separate folders and each of them has one composer.lock file. 
Also i have large php projects where i don't use composer so all files are parsed on start. It would be nice to manually specify which should be cached.

catroll:
Ubuntu 17.10
VS Code 1.19.1
PHP Intellisense 2.2.2
Memory Size: 16G, almost used by a process called "PHP Language Server"

```
$ ps -ef | grep -E "9099|6596"
catroll   6596  6541  1 09:50 ?        00:00:07 /usr/share/code/code --nolazy --inspect=9333 /usr/share/code/resources/app/out/bootstrap --type=extensionHost
catroll   6629  6596  1 09:50 ?        00:00:04 /usr/share/code/code /usr/share/code/resources/app/extensions/html/node_modules/vscode-languageclient/lib/utils/electronForkStart /usr/share/code/resources/app/extensions/html/server/out/htmlServerMain.js --node-ipc --clientProcessId=6596
catroll   9302  6596 99 09:56 ?        00:00:19 PHP Language Server
[...]

PR #641, opened 2018-05-10
Ensure all dependencies in a multi-root project are cached

dmakovec:
I've had issues with a multi-root project where CPU runs through the roof every time https://github.com/felixfbecker/vscode-php-intellisense is opened.

The reason for this was LanguageServer.php was arbitrarily choosing the first composer.lock that it came across within rootPath, then indexing and caching dependencies from it alone.  Consequently, whenever VSCode restarts, php-language-server re-indexes all of the vendor packages in sub-folders that weren't associated with the selected composer.json/lock file.

My quick patch here simply iterates over all composer.lock files within the multi-root project, indexes then caches to disk every single dependency it uncovers. The result of this being that subsequent startup times are much faster and CPU no longer tanks.

This is my simple solution which I believe partially solves what https://github.com/felixfbecker/php-language-server/pull/509 is attempting to address. Possibly related to https://github.com/felixfbecker/php-language-server/issues/465

### DIFF
--- a/src/LanguageServer.php
+++ b/src/LanguageServer.php
@@ -215,10 +215,18 @@ class LanguageServer extends AdvancedJsonRpc\Dispatcher
                     sortUrisLevelOrder($composerLockFiles);
 
                     if (!empty($composerLockFiles)) {
-                        $this->composerLock = json_decode(yield $this->contentRetriever->retrieve($composerLockFiles[0]));
+                        $packages = [];
+                        $packagesDev = [];
+                        foreach($composerLockFiles as $composerLockFile) {
+                            $composerLock = json_decode(yield $this->contentRetriever->retrieve($composerLockFile));
+                            $packages = array_merge($packages, $composerLock->packages);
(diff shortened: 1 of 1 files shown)


src/LanguageServer.php
```diff
@@ -215,10 +215,18 @@ public function initialize(ClientCapabilities $capabilities, string $rootPath =
...
+                        $packages = [];
+                        $packagesDev = [];
+                        foreach ($composerLockFiles as $composerLockFile) {
+                            $composerLock = json_decode(yield $this->contentRetriever->retrieve($composerLockFile));
+                            $packages = array_merge($packages, $composerLock->packages);
+                            $packagesDev = array_merge($packagesDev, $composerLock->{'packages-dev'});
+                        }
+                        $this->composerLock = $composerLock;
+                        $this->composerLock->packages = $packages;
+                        $this->composerLock->{'packages-dev'} = $packagesDev;
```
felixfbecker: Hmm, are you sure aggregating all packages work? What if the same package appears with different versions, and then the cache key that is used for the index is not actually the package version that was indexed?

PR #661, opened 2018-07-29
Use protocol package

dantleech:
This PR adapts the Language Server to use the extracted [php language server protocol](https://github.com/felixfbecker/php-language-server-protocol) package.

Change summary:

- [x] Removed the `Protocol` namespace.
- [x] Introduced a `Factory` namespace containing the static factory methods removed from the extracted packages.
- [x] Moved the `Protocol/Message` class to the root namespace (it doesn't seem tobe [part of the protocol](https://github.com/felixfbecker/php-language-server-protocol/pull/1#discussion_r205975220)).

Not too sure how well the `Factory` namespace fits into the architecture. As all the factories produce Protocol objects, might make more sense to put them in `Protocol` or `Protocol\Factory`.

TODO:

- [ ] ...


composer.json
```diff
@@ -14,6 +14,12 @@
         "autocompletion",
         "refactor"
     ],
+    "repositories": [
+        {
+            "type": "vcs",
+            "url": "https://github.com/dantleech/php-language-server-protocol"
+        }
+    ],
```
dantleech: If this PR seems OK, then we can add a `felixfbecker/language-server-protocol` package to packagist?

composer.json
```diff
@@ -24,6 +30,7 @@
         "php": "^7.0",
         "composer/xdebug-handler": "^1.0",
         "felixfbecker/advanced-json-rpc": "^3.0.0",
+        "felixfbecker/language-server-protool": "*",
```
dantleech: Should we start on `1.0.0` or `0.1.0` ?

src/Factory/SymbolInformationFactory.php
```diff
@@ -1,54 +1,26 @@
...
-
     /**
      * Converts a Node to a SymbolInformation
      *
      * @param Node $node
      * @param string $fqn If given, $containerName will be extracted from it
      * @return SymbolInformation|null
      */
-    public static function fromNode($node, string $fqn = null)
+    public static function fromNode($node, string $fqn = null):? SymbolInformation
```
felixfbecker: We need to support PHP 7.0, which doesn't have nullable types

felixfbecker:
- I like the `LanguageServer\ProtocolBridge` namespace, I think that would be a good place to put all the factories.
- The `Message` class should be in the protocol package because it is specced as part of the protocol: https://github.com/Microsoft/language-server-protocol/blob/gh-pages/specification.md#base-protocol

dantleech:
> The Message class should be in the protocol package because it is specced as part of the protocol

It is, but it does not conform to the [type according to the protocol](https://github.com/Microsoft/language-server-protocol/blob/gh-pages/specification.md#abstract-message) (which only has a `$jsonrpc` property).

I thought it would be better to keep this an implementation detail and implement the version indicated in the spec in the protocol (even if it's not used here). Does that make sense?

felixfbecker:
I think you are misunderstanding the class - the LSP `Message` class represents an LSP message consisting of headers and a body as defined by the LSP spec, saved in `$headers` and `$body`. `$body` then is the [JSON RPC `Message`](https://github.com/felixfbecker/php-advanced-json-rpc/blob/master/lib/Message.php) object, which has the `$jsonrpc` property.

dantleech:
@felixfbecker what do you think re. `Message`? having another look it also seems to sit uncomfortably in the same namespace as f.e. `MessageActionItem` and `MessageType` - which have nothing to do with our `Message` here.

Thanks for your time by the way :)

felixfbecker:
Maybe have sub-namespaces `Types` and `Transport`?

dantleech:
So:

```
Types\
    ClientCapabilities
    CodeActionContext
    CodeLens
    CodeLensOptions
    ...
Transport\
   Message
```

Although putting `Message` in `Transport` and leaving the others where they are also makes sense to me.

felixfbecker:
Yes, looks good to me

felixfbecker:
@dantleech do you plan to bring this over the finish line?

dantleech:
@felixfbecker yeah, sorry have been working on other things, but it's still on my roadmap, hopefully this coming weekend.

dantleech:
hmm, looking at this again, the `Message` class really is quite coupled to `AdvancedJsonRpc`.

For example:

```php
// ClientHandlerTest.php
$reader->write(new Message(new AdvancedJsonRpc\SuccessResponse($msg->body->id, 'pong')));
```

(it is above expected that the `body` property is an AdvancedRPC object)

I don't know if we want to create `Request`, `Response`, etc DTOs mirroring the data structures in AdvancedJsonRpc? I still would vote for leaving the `Message` here where it can be coupled to that library though.

dantleech:
I've updated to the latest version of the `protocol` PR, which includes the namespace-renaming BC break.

felixfbecker:
Published the protocol package. Should I release a 1.0.0?

dantleech:
> Should I release a 1.0.0?

I've pushed an update to use the published package, if the builds pass then 1.0.0 sounds good to me.

felixfbecker:
Released v1.0.0

dantleech:
Updated

composer.json
```diff
@@ -24,6 +24,7 @@
         "php": "^7.0",
         "composer/xdebug-handler": "^1.0",
         "felixfbecker/advanced-json-rpc": "^3.0.0",
+        "felixfbecker/language-server-protocol": "1.0.0",
```
felixfbecker: Should be `^1.0.0`

Issue #548, opened 2017-12-08
Configurable file size limit (for laravel-ide-helper)


jwhulette:
I'm using https://github.com/barryvdh/laravel-ide-helper but I get this message when indexing, Ignoring file.../_ide_helper.php because it exceeds size limit of 150000 bytes (539143).

Can I increase the file limit size?

str:
Usually large files are the "let's put everything there" solution. I hate it, but it is what it is. We need to be able to set the size limit in the IDE configuration so we can include those ugly files.

robberphex:
Maybe index a large file will cost a lot of time, but can we use some extension of PHP to accurate indexing? [php-ast](https://github.com/nikic/php-ast)

---
P.S.
*I am working on a project with 17921 `*.php` files, the largest file `Types.php`, 968KiB. I really want:*
1. accelerate the indexing.(via [php-ast](https://github.com/nikic/php-ast)?)
2. language server can index large files.
3. `Go to implements`

AndisGrossteins:
I have two questions for developer of `php-language-server`:
1. WTF?
2. Why hardcode the limit in code?

https://github.com/felixfbecker/php-language-server/blob/26e3451e6123aff9e40c8465d5d7d013ba2a2724/src/PhpDocumentLoader.php#L109

```php
    public function load(string $uri): Promise
    {
        return coroutine(function () use ($uri) {
            $limit = 150000;
            $content = yield $this->contentRetriever->retrieve($uri);
            $size = strlen($content);
            if ($size > $limit) {
                throw new ContentTooLargeException($uri, $size, $limit);
            }
            if (isset($this->documents[$uri])) {
                $document = $this->documents[$uri];
                $document->updateContent($content);
            } else {
                $document = $this->create($uri, $content);
            }
            return $document;
        });
    }
```

RobberPhex:
Excuse me, I'm not working on this, so I will close this PR.

animalillo:
An option to configure this would be great, i have to deal with some huge files and are a pain to work with without intellisense working properly.

phil-s:
Is this hard-coded limit connected to issue #245 ?

phil-s:
I have nine files above the 150,000 byte limit in my project:

* 201,651
* 203,613
* 208,023
* 209,602
* 235,468
* 319,514
* 471,662
* 553,415
* 577,243

[...]

Issue #299, opened 2017-02-14
Allow configurable file size limit


cantonese84:
The current fixed file size limit of 150000 is too small.  Drupal 7's modules/system/system.api.php file size is 200K.  It would be best if this is configurable (i.e. by vscode-php-intellisense).

felixfbecker:
Feel free to do a PR

Issue #302, opened 2017-02-15
[Feature Request] Add validator frequency setting


aidantwoods:
As discussed in #301, vanilla VS Code allows setting either `"php.validate.run": "onSave"` or `"php.validate.run": "onType"`.

Current behaviour replicates the setting for `"php.validate.run": "onType"`, feature request to add an option to choose between `"onType"` and `"onSave"` behaviour.

felixfbecker:
PR welcome

PR #668, opened 2018-08-29
Feature: configure file size limit

JSteitz:
Requires the PR from vscode-php-intellisense felixfbecker/vscode-php-intellisense#77
This PR is based on #308 and should be merged after

This solves issues / feature requests for #299 #548


src/Indexer.php
```diff
@@ -221,7 +221,7 @@ private function indexFiles(array $files): Promise
                 yield timeout();
                 $this->client->window->logMessage(MessageType::LOG, "Parsing $uri");
                 try {
-                    $document = yield $this->documentLoader->load($uri);
+                    $document = yield $this->documentLoader->load($uri, $this->options->fileSizeLimit);
```
JSteitz: @felixfbecker I think this is the right place to pass the file size limit, what do you think?
Just asking, because I need to adjust a lot of tests for that.

felixfbecker:
> because I have difficulties to understand why clients can't implement the initialization options. 

They _can_. But it's not defined in the protocol that `initializationOptions` contains _configuration_. So you basically have to tell _every_ client "if you want this to work, you need to do this special thing that is not defined in the protocol". That is against the purpose of LSP - the client should not have to do special things per language server. And in this case, there is a _proper_ way defined in the protocol to handle confiuration, so we should use that.

JSteitz:
Alright, now I get it :) thx for the patience.

Here is my suggestion that should satisfy LSP:

We can hook into `initialized` and request configuration settings with `workspace/configuration`.
This also means that the indexer must be moved here. This shouldn't be a problem, since it is directly called after `initialize`. With this approach we can target only settings that will affect the indexer and use the `didChangeConfiguration` for other settings (e.g. Add validator frequency setting #302)

I really really don't want to start reindexing after every change in `didChangeConfiguration`.
We also can add a command to invalidate cache and reindex project.

felixfbecker:
Yeah, using `workspace/configuration` sounds fine to me, I think that's actually a recent addition that I didn't hear about before.

> I really really don't want to start reindexing after every change in didChangeConfiguration.

It would only have to reindex if a setting affecting the indexing changed, e.g. the file size limit. What is the problem with that?

JSteitz:
Yes that is right, that it would only reindex affected settings.
See https://github.com/felixfbecker/php-language-server/pull/308#discussion_r101897846

If we would move it to `didChangeConfiguration` then we need the ability to abort current running indexing (can we?) and restart with the new settings.
Personally I don't care if we can make it work like that.
If we can't than I don't want to index 2 times.

felixfbecker:
We could pretty easily. Ideally we would pass some sort of `CancellationSignal` to coroutines. But a simple implementation can also just be letting the indexer have a private property that gets incremented with each run, and on every iteration, if the property changed, abort the current run.

JSteitz:
Alright, which of these 2 Options should I implement.

1. `initialized` - reindexing after restart
2. `didChangeConfiguration` - reindexing without restart

felixfbecker:
Both sounds like the best way - so we only start indexing after we know the initial configuration, but are still able to handle configuration changes (a `didChangeConfiguration` call without relevant changes should not trigger a reindex).

JSteitz:
After testing with `workspace/configuration` and `initialized` I came to the conclusion, that we can not use that for now. It seems like it requires multiple workspaces to be active to work (capabilities for that are not set and a request results in "unhandled method ...").

So only the option for `didChangeConfiguration` is remaining.
Can you hint me how I can implement an CancelationSignal?
Otherwise I will do a simple method to cancel the indexing process.

felixfbecker:
> It seems like it requires multiple workspaces to be active to work (capabilities for that are not set and a request results in "unhandled method ...").

Could you explain this in more detail?

> So only the option for didChangeConfiguration is remaining.
> Can you hint me how I can implement an CancelationSignal?
> Otherwise I will do a simple method to cancel the indexing process.

I would go with the simple method for now

JSteitz:
Now I feel dump...
Thx for the help, after checking everything again so I don't tell bullshit I saw that the languageclient version was wrong for this feature.

Did an upgrade and now I see it :)
Will work on this later again.

felixfbecker:
@JSteitz are still interested in getting this merged?

PR #670, opened 2018-08-31
Fix infinite loop when indexing self referencing classes

d0x2f:
As the indexer attempts to parse the parent class, it adds itself to the array being iterated.

A simple check that the FQN isn't identical resolves the issue.

### DIFF
--- a/fixtures/self_referencing_class.php
+++ b/fixtures/self_referencing_class.php
@@ -1,0 +1,4 @@
+<?php
+class C extends C {}
+$c = new C;
+$c->undef_prop = 1;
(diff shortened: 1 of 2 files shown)


felixfbecker:
What about a loop like this?
```php
class A extends C {}
class B extends A {}
```

d0x2f:
indeed that causes the same looping issue.

I'm not familiar enough with the codebase to know if there's a list of already processed FQN's we can lookup.

I'll think on it.

d0x2f:
I've added a list of visited fqns and changed the check to prevent adding a parent fqn that's already been visited.

Let me know what you think.

PR #679, opened 2018-10-28
Support inherited static functions

letynsoft:
When the resolveScopedPropertyAccessExpressionNodeToFqn cannot find the definition, try it the extended class


letynsoft:
Hmm, from the tests it really doesn't like that, but from the functionality it works nice :-(
Is there any way to do this only when this is request from the editor? 
They doesn't seem to be added for a right reason in the tests...

felixfbecker:
Could you explain what this PR fixes? What do you mean with parent contexts?

letynsoft:
eg.
```
class a {
  static function a () { ... }
}
class b {
}
```
and then when you call ```b::a()``` the language server is able to find the method then (same as PHP does)

src/DefinitionResolver.php
```diff
@@ -493,21 +493,37 @@ private function resolveScopedPropertyAccessExpressionNodeToFqn(Node\Expression\
...
+                    return null;
+                }
+                $memberName = $scoped->memberName->getName();
+                if (empty($memberName)) {
+                    return null;
+                }
+                $name = (string)$className . '::$' . $memberName;
+            } else {
+                $name = (string)$className . '::' . $scoped->memberName->getText($scoped->getFileContents());
+            }
```
felixfbecker: There is a lot of logic here that does not change on each loop iteration, such as the `instanceof` checks above and the member name. Could you save this in variables and move it out of the loop?

felixfbecker (changes requested):

felixfbecker:
Also, could you add a test?

letynsoft:
Hi,
The standard TestClass method and variable finds the reference to the call/access, so i think that is what i can test there :)

PR #682, opened 2018-11-10
Don't suggest private methods

### LINKED ISSUE #469: Intellisense autocomplete also suggests private or protected method and properties
Just like said in the title it also suggest the methods that is declared private/public

image

Here is the class implementing it
image

Although `php` will also throw an error for this might be better if the *intellisense* won't suggest anymore.

gnoe:
To avoid errors on Travis with line ending.

This branch has the same fix for showing private and protected methods and properties out of the context.

Fixes #469


gnoe:
Yes, I can.
Should I add it to tests\Server\TextDocument\CompletionTest.php? 
I tested this launching the php-intellisense extension pointing to my local copy of the language server.
I will write the test.

Travis keeps failing because of the line ending.

felixfbecker:
What editor are you using? You should configure it to trim trailing whitespace (or delete it manually)

gnoe:
I use VS Code and Netbeans, I configured VS Code.
Thanks!
Next commit with the test should be fixed

gnoe:
I did merge the branch already, sorry, I have to get used to the code. I removed the part of the static checking. I would like to test it better.

PR #690, opened 2018-11-28
Add a phan config to analyze this project

TysonAndre:
https://github.com/felixfbecker/php-language-server/pull/612#issuecomment-442093319

The inline var plugin was forked to remove `:void`.
It's also possible to depend on
https://packagist.org/packages/drenso/phan-extensions
and just run phan with php 7.2
(or to request support for php 7.0 in a release)

- inline vars are used to be much more accurate in detecting unused
  `use` statements.

  Phan does not use those annotations for type inference.

This PR does not attempt to fix any of the detected issues.
You may wish to uncomment the suppression in `.phan/config.php`,
re-run analysis, and fix all of the issues in follow-up PRs
(or move from project suppressions to more granular suppressions)

- https://github.com/phan/phan/wiki/Tutorial-for-Analyzing-a-Large-Sloppy-Code-Base
  has tips on strengthening analysis
- Issues can be suppressed at various granularities
  (project, file, structural element, and line of code). See
  https://github.com/phan/phan/wiki/Annotating-Your-Source-Code#suppress and
  https://github.com/phan/phan/wiki/Annotating-Your-Source-Code#phan-suppress-current-line
- There are various config settings that can be tuned - common ones are
  mentioned in the wiki.

---

If 1 or more (unsuppressed) issues is emitted, then the `vendor/bin/phan` step will have a non-zero exit code, and fail the Travis CI build

phan/phan's version is pinned so that a new release of Phan won't cause this project's build to fail (e.g. if it gets better at detecting issues)


.travis.yml
```diff
@@ -18,6 +18,8 @@ install:
   - composer install --prefer-dist --no-interaction
 script:
   - vendor/bin/phpcs -n
+  - pecl install ast-1.0.0
```
felixfbecker: This should be in the `install` section

.phan/plugins/InlineVarPlugin.php
```diff
@@ -0,0 +1,203 @@
+<?php
```
felixfbecker: Can we avoid forking this here? I would prefer any of the other options you mentioned than maintain this

TysonAndre:
Revisiting this, it looks like the inline var plugin isn't needed any more. I forget why I had it in my local checkout of php-language-server - maybe this project used more inline var annotations in the past, or maybe I copied the .phan/config.php from another project I was analyzing.

Uncommenting the UnreferencedUseNormal suppression, I saw the exact same issues with/without the suppression.

Aside: If you need the inline var plugin in the future, you can use something along these lines to only run php in one of the jobs (or you could check PHP_VERSION_ID >= 70100)

```
install:
  - composer install --prefer-dist --no-interaction
  - if [[ "$(php -r 'echo PHP_MAJOR_VERSION.PHP_MINOR_VERSION;')" == 72 ]]; then pecl install ast-1.0.0; composer.phar require --dev drenso/phan-extensions=2.4.1; fi

script:
  - vendor/bin/phpcs -n
  - if [[ "$(php -r 'echo PHP_MAJOR_VERSION.PHP_MINOR_VERSION;')" == 72 ]]; then vendor/bin/phan --plugin vendor/drenso/phan-extensions/Plugin/DocComment/InlineVarPlugin.php; fi
  - vendor/bin/phpunit --coverage-clover=coverage.xml --colors=always
  - bash <(curl -s https://codecov.io/bash)
```

.phan/plugins/InlineVarPlugin.php
TysonAndre: Removed - this seems unnecessary

.travis.yml
TysonAndre: Done

felixfbecker:
I'm surprised there are no complaints by Phan!

PR #694, opened 2018-12-12
Run 'composer install' in a docker builder stage

jkblume:
I wondered on the comment in the Dockerfile, which says to install the composer dependencies before building the image. 

Maybe it is an option for you to use a multi-stage Dockerfile, like i proposed in my PR :) I would appreciate to hear your thoughts about that.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,17 +1,21 @@
 
+FROM composer AS builder
+
+COPY ./ /app
+RUN composer install
+
 # Running this container will start a language server that listens for TCP connections on port 2088
 # Every connection will be run in a forked child process
 
 # Please note that before building the image, you have to install dependencies with `composer install`
 
 FROM php:7-cli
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -1,17 +1,20 @@
 
+FROM composer AS builder
+
+COPY ./ /app
+RUN composer install
+
 # Running this container will start a language server that listens for TCP connections on port 2088
 # Every connection will be run in a forked child process
```
felixfbecker: Can you keep this comment at the top?

felixfbecker (commented):
Nice!

Dockerfile
jkblume: Yes of course :)

jkblume: Moved the comments back to the top. I also replaced the deprecated maintainer with the appropriate "maintainer" label.

felixfbecker (approved):

PR #746, opened 2019-06-19
migrate to amphp

robberphex:


felixfbecker:
Thanks for this! Could you give some explanation for this change in the PR description? I.e. what is the motivation/benefit, what are the high-level required changes this PR implements, are there any drawbacks?

PR #754, opened 2019-09-21
Capture initialized notification

joas8211:
Initialized notification is specified in LSP but wasn't implemented. Without implementation JSON-RPC would throw a method not found error.

https://microsoft.github.io/language-server-protocol/specification#initialized


joas8211:
Wrong branch...

joas8211:
#755 is the correct PR

PR #756, opened 2019-09-23
phpDocumentor may throw RuntimeException

Ciantic:
Many places in the phpDocumenter may throw RuntimeException, naturally it's hard to tell what went wrong because so many places is coded with that.


Ciantic:
This is appearing again, please consider merging this.

Issue #21, opened 2016-09-09
Rename


felixfbecker:
https://github.com/Microsoft/language-server-protocol/blob/master/protocol.md#rename-request

still-dreaming-1:
[![Bountysource](https://api.bountysource.com/badge/issue?issue_id=37715880)](https://www.bountysource.com/issues/37715880-rename?utm_source=37715880&utm_medium=shield&utm_campaign=ISSUE_BADGE) Other people can also add to this bounty if they want.

echosa:
The spec documentation URL in the ticket description is out of date. This seems to be the new correct URL: 

https://microsoft.github.io/language-server-protocol/specification#textDocument_rename

erlangparasu:
+1

matt-allan:
I have this partially working on a branch.  I don't have time at the moment to finish it.  Maybe it will work as a good starting point for someone else?

https://github.com/matt-allan/php-language-server/tree/rename-functionality

PR #768, opened 2019-11-12
Rename support

rummatee:
This PR attempts to implement the rename function. Therefore it implements #21 .
There are however still some limitations:
If there isn't a reference in the same file, as the definition, the definition doesn't get changed.
Also at least in my editor (nvim) it is only possible to do one rename and it's necessary to restart before doing the next rename.
I would appreciate any help with these issues or any further comments.

### DIFF
--- a/src/Index/Index.php
+++ b/src/Index/Index.php
@@ -187,6 +187,9 @@ class Index implements ReadableIndex, \Serializable
     public function removeDefinition(string $fqn)
     {
         $parts = $this->splitFqn($fqn);
+        if (empty($parts)) {
+            throw new \Exception($fqn);
+        }
         $this->removeIndexedDefinition(0, $parts, $this->definitions, $this->definitions);
 
         unset($this->references[$fqn]);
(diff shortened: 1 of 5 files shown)


felixfbecker (changes requested):
Thanks for taking a stab at this! However, the issues in the description seem like blockers before merging this.

rummatee:
I rewrote the definition finding and added code to update the index after the rename. The issues I described earlier should be solved with this.

src/Index/Index.php
```diff
@@ -423,6 +426,9 @@ private function indexDefinition(int $level, array $parts, array &$storage, Defi
      */
     private function removeIndexedDefinition(int $level, array $parts, array &$storage, array &$rootStorage)
     {
+        if (empty($parts)) {
```
rummatee, line 426: I don't really understand this function well enough to be confident about this code, but it fixes an exception that I got before and it seems intuitively right to have a base case for the recursion.

Issue #795, opened 2020-11-02
LanguageServer script failing


phatwhalemedia:
Guys/ladies, I'm having some trouble getting composer/language-server setup in Ubuntu Linux. This is the latest error in my quest to finish this apparently insurmountably difficult task. Granted I'm new to composer, so this is all still a bit confusing.

After running

`composer run-script --working-dir=vendor/felixfbecker/language-server parse-stubs
`

The script fails on line

`Parsing file:///home/phatwhalemedia/.config/composer/vendor/jetbrains/phpstorm-stubs/http/http.php
`

Giving this error

```
Script LanguageServer\ComposerScripts::parseStubs handling the parse-stubs event terminated with an exception

  [RuntimeException]                         
  A type is missing before a type separator
```

If there's any further info required, please let me know.

I've been following the steps from the following link:
#611

pfernandez:
Same. My "workaround" was to simply remove `~/.config/composer/vendor/jetbrains/phpstorm-stubs/http/http.php` for the time being.

3ynm:
same err here :-1:

apfelchips:
You can use the bundled language server from the vscode extension https://marketplace.visualstudio.com/items?itemName=felixfbecker.php-intellisense --> download extension --> rename to .zip --> /extension/vendor/felixfbecker/language-server/bin

AJenbo:
Also fails on vendor/jetbrains/phpstorm-stubs/standard/standard_8.php

CodyWilson:
I submitted a simple pull request to fix this.

The simple solution until that's merged is to setup your composer.json like so:
```json
{
    "repositories": [
        {
            "type": "vcs",
            "url": "https://github.com/CodyWilson/php-language-server"
        }
    ],
    "require": {
        "felixfbecker/language-server": "dev-master"
    }
}
```

Alternatively you could fork it yourself and do that change, and set the version control to point to your fork.

Mte90:
@felixfbecker any updates for this project?

PR #797, opened 2020-12-07
Fixing build by setting stubs ver. to older ver.

CodyWilson:
phpstorm-stubs dev-master already had PHP8 features which are breaking
the parser per #795.
The simple solution for this is to force an older version of the stubs.


CodyWilson:
As a note, the appveyor build failing appears to be due to updates to `microsoft/tolerant-php-parser`.
Which also isn't really controlled well for versions, then again they're just sticking all changes in `0.0` and not iterating for anything that could be considered a breaking change.

